PR #7, opened 2025-02-05
Preliminary PR to merge my fork into a proper branch, as requested

Patrigan:
Compiling Runegems, with enchantment backed modifiers. Preliminary, untiered, unrolled implementation of Sockets.


Attack8:
I'm leaving this because it's funny, will fix!

Issue #17, opened 2025-02-09
Inventory Recovery Death Mechanics


McMelonTV:

immortius:
- [ ] Basic Snapshot capture and restore
- [ ] Support overly full inventory on restore
- [ ] Track entity blocks that are placed and recovered (e.g. Shulkers)
- [ ] Handle container items
- [ ] Handle ender chests
- [ ] BUCKETS

immortius:
Making the decision to skip ender chests for the PoC. It is likely that we'll want to not support them being used in rifts due to their interaction with other mods (specifically could be used to egress loot from the player's inventory into the overworld). It wouldn't be difficult to extend the system to support them otherwise.

PR #25, opened 2025-02-11
Inventory snapshot system

immortius:
Issue #17 

### Summary

This pull request introduces a proof of concept inventory snapshot and restoration on death mechanic. This system allows a snapshot of they player's inventory to be captured. When they next die the items they still have on them (technically, are about to drop as per the LivingDropsEvent) are compared to the snapshot, and the items sorted between those they should retain and those that should be dropped. Those that should be retained are then restored to the player when they respawn.

### Usage instructions

Run the command `dimensiondelvers:createInventorySnapshot` to save current inventory to a snapshot. Restoration is triggered upon next death and respawn.

### Implementation notes

- Non-stackable items are given a Inventory Snapshot Id data component (a UUID) which is used to track them.
- InventorySnapshots are stored as an attachment on the player. These include both a list of ids, and a list of all stackable items (as a copy of the original item stacks).
- A RetainInventorySnapshotLootModifier is used to ensure Inventory Snapshot Ids survives for items that are placed as a block and then picked up.
- Items that use the Container data component are supported. The contents of such items are captured in snapshots, and are also reviewed upon death.

### Known issues

 - Filled buckets lose their ids when emptied, so will not be retained on death if emptied and refilled.
 - Ender chests are not currently part of inventory snapshots. This would be easy to add, but suspect we will want to disable them in rifts due to mod interactions.
 - Mod containers have not been catered for. Likely will need to add a way for registering container support with the system for different container types.
 - Potentially LootModifiers should be switch to use datagen, if we have further use for them at least.

- Extension support for new types of containers
- Loot Modifier could likely be improved (maybe create a LootCondition for BlockEntities?)
- Could change snapshots from holding item stacks for stackable items, to holding a mapping of item to count. This would better support large quantities of the same item (most likely from modded containers)
- Handling of dropped items will need to change when an option or options for handling them is decided. Perhaps could produce an event for them.
- Command will need to be dropped once integrated with rifts, or else improved and made an op command for debug testing.

### DIFF
--- a/src/main/java/com/dimensiondelvers/dimensiondelvers/server/inventorySnapshot/InventorySnapshotSystem.java
+++ b/src/main/java/com/dimensiondelvers/dimensiondelvers/server/inventorySnapshot/InventorySnapshotSystem.java
@@ -1,0 +1,217 @@
+package com.dimensiondelvers.dimensiondelvers.server.inventorySnapshot;
+
+import com.dimensiondelvers.dimensiondelvers.init.ModAttachments;
+import com.dimensiondelvers.dimensiondelvers.init.ModDataComponentType;
+import net.minecraft.core.component.DataComponentPatch;
+import net.minecraft.core.component.DataComponents;
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.world.entity.item.ItemEntity;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.component.ItemContainerContents;
+import net.neoforged.neoforge.event.entity.living.LivingDropsEvent;
+
+import java.util.*;
+
+/**
+ * Implementation for capture, update and restoration of Inventory Snapshots
+ * <p></p>
+ * The envisioned behavior is:
+ * <ul>
+ *     <li>When a snapshot is initially captured, all items in the players inventory and sub-inventories are enumerated</li>
+ *     <li>When a player dies, the player's inventory is compared to the snapshot. The snapshot is reduced to the remaining items and any excess is dropped (mechanism TBD)</li>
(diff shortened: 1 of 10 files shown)


iwolfking:
Looks awesome, I will try to review this more soon and get it merged so we can start to play around with this if no one else has any issues.

We will almost certainly want to deal with restoring Sophisticated Backpacks to their original state if that is something you wanted to look into adding to this, I know you made a note about not supporting modded containers currently. That would be the primary one we would need to support.

immortius:
> We will almost certainly want to deal with restoring Sophisticated Backpacks to their original state if that is something you wanted to look into adding to this, I know you made a note about not supporting modded containers currently. That would be the primary one we would need to support.

Sure, I can move on to investigating that.

src/main/java/com/dimensiondelvers/dimensiondelvers/server/inventorySnapshot/InventorySnapshotSystem.java
```diff
@@ -0,0 +1,210 @@
...
+                } else if (item.isStackable()) {
+                    int dropCount = calculateDropCount(item);
+
+                    if (retainingContainer && dropCount > 0) {
+                        dropItems.add(createItemEntity(item.split(dropCount)));
+                    } else if (!retainingContainer && dropCount < item.getCount()) {
+                        retainItems.add(item.split(item.getCount() - dropCount));
+                    }
+                } else if (retainingContainer) {
+                    item.applyComponents(REMOVE_SNAPSHOT_ID_PATCH);
```
immortius, line 160: There is a missing check for if the item is a container and processing of its contents here.

immortius:
Closed and created #25 with updated code line with mod container extension support and various other changes.

PR #27, opened 2025-02-11
Create Modifier System

Rev219:
Created modifier system

NOTE: "src/main/java/com/dimensiondelvers/dimensiondelvers/modifier/Modifier" is still being worked on


src/main/java/com/dimensiondelvers/dimensiondelvers/modifier/Modifier Ideas.md
Patrigan, line 1: This file can be deleted, I think?

Patrigan (approved):
Aside from the .md file, everyone seems to be there.

I might want to see the ModModifierEffects split into multiple files. Maybe moving all the Resource Keys into one file? Unsure, to be discussed.

PR #28, opened 2025-02-14
Feature: Processor blocks

Patrigan:
This PR adds two things:

A set of code to easily register and add a whole Block Family.

4 Block families for Processor Blocks.


src/main/java/com/dimensiondelvers/dimensiondelvers/block/BlockFamilyHelper.java
```diff
@@ -0,0 +1,165 @@
+package com.dimensiondelvers.dimensiondelvers.block;
+
+import net.minecraft.data.BlockFamily;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.level.block.Block;
+
+import java.util.function.Supplier;
+
+import static net.neoforged.neoforge.internal.versions.neoforge.NeoForgeVersion.MOD_ID;
```
immortius, line 9: Is this the MOD_ID constant you intended? I think ours is DimensionDelvers.MODID confusingly

src/main/java/com/dimensiondelvers/dimensiondelvers/block/BlockFamilyHelper.java
```diff
@@ -0,0 +1,165 @@
...
+            return this;
+        }
+
+        public Builder withTrapdoor(Supplier<Block> trapdoor) {
+            this.trapdoor = trapdoor;
+            return this;
+        }
+
+        public BlockFamilyHelper createBuildBlockHelper() {
+            return new BlockFamilyHelper(blockId, block, slab, stairs, button, pressurePlate, wall, fence, trapdoor);
```
immortius, line 160: It might be worth adding a check here that all the values have been populated, just to fail early in case one was missed. Although it will be clear eventually when a NPE is raised.

src/main/java/com/dimensiondelvers/dimensiondelvers/block/BlockFamilyHelper.java
Patrigan: Dumb. Fixed it.

src/main/java/com/dimensiondelvers/dimensiondelvers/datagen/ModLanguageProvider.java
```diff
@@ -28,10 +31,26 @@ protected void addTranslations() {
...
         // Adds a generic translation
         add("itemGroup.dimensiondelvers", "Dimension Delvers");
 
         add("container.dimensiondelvers.rune_anvil", "Rune Anvil");
     }
+
+    private static @NotNull String getTranslationString(Block block) {
+        String idString = BuiltInRegistries.BLOCK.getKey(block).getPath();
+        StringBuilder sb = new StringBuilder();
+        for (String word : idString.toLowerCase().split("_")) {
```
immortius, line 49: I recommend specifying a Locale when using toLowerCase (Probably `Locale.ROOT`).

immortius: Same for toUpperCase below

Patrigan: Done

immortius (approved):
Everything I have raised has been addressed. I note merging makes further review difficult and would encourage rebasing but that is a discussion for another place :)

Patrigan:
Sorry Immortius, had to do a merge> Rebase was giving too many conflicts.

PR #29, opened 2025-02-15
feat: #17 Initial implementation of inventory snapshot

immortius:
### Summary
This pull request introduces a proof of concept inventory snapshot and restoration on death mechanic. This system allows a snapshot of they player's inventory to be captured. When they next die the items they still have on them (technically, are about to drop as per the LivingDropsEvent) are compared to the snapshot, and the items sorted between those they should retain and those that should be dropped. Those that should be retained are then restored to the player when they respawn.

### Usage instructions
Run the command `dimensiondelvers:createInventorySnapshot` to save current inventory to a snapshot. Restoration is triggered upon next death and respawn.

### Implementation notes
* A snapshot is a combination of a UUID and a list of item stacks.
* The UUID is attached as a data component to "non-stackable items" a player has, recursively through their inventory and any container items
* A copy of all stackable items is added to the list of item stacks.
* The snapshot is stored as an attachment on the player. 
* A RetainInventorySnapshotLootModifier is used to ensure Inventory Snapshot Ids survives for items that are placed as a block and then picked up.
* ContainerTypes can be registered with the snapshot system to add support for additional container types (e.g. modded backpacks). An implementation for items that use the Container data component (e.g. Shulker boxes) is registered by default.

### Known issues
* Filled buckets lose their ids when emptied, so will not be retained on death if emptied and refilled.
* Ender chests are not currently part of inventory snapshots. This would be easy to add, but suspect we will want to disable them in rifts due to mod interactions.
* Potentially LootModifiers should be switch to use datagen, if we have further use for them at least.

* Loot Modifier could likely be improved (maybe create a LootCondition for BlockEntities?)
* Could add a system for specifying which items are "stackable" so that we can better control what is handled via data component tagging and what is captured as item stacks.
* Handling of retained and dropped items will need to change when an option or options for handling them is decided. Perhaps could produce an event for them.
* Command will need to be dropped once integrated with rifts, or else improved and made an op command for debug testing.


iwolfking:
Thank you! I will review this tomorrow after our meeting probably.

iwolfking (approved):
This looks good, only thing is that server/inventorySnapshot package is a bit messy, could be be a bit more organized, but we can do some refactoring later.

PR #45, opened 2025-02-17
Runtime mods

Patrigan:
Just a branch to add a couple runtime mods to make modding easier.


Anorak01 (approved):
Looks good to me
*praying to Gradle gods it actually is good

PR #49, opened 2025-02-21
Prevent durability damage in its entirety

a1qs:
This PR removes durability damage from the game in its entirety, by always returning false when the `ItemStack#isDamageableItem` method is called.


Patrigan (approved):
Not the biggest fan of mixins. I wonder if we can not play around with default datacomponents, like unbreakable. But this will do for now..

PR #50, opened 2025-02-24
PoC Rift Entrance and Basic Key

immortius:
### Summary
This pull request introduces a proof of concept Rift Key item, Rift Spawner block and Rift Entrance entity.

### Usage instructions
When a Rift Key is used on a placed Rift Spawner block, a Rift Entrance entity will be created. If there is three non-colliding block spaces above the Rift Spawner block, the Rift Entrance will be created above it and will rotate to face the player. Otherwise if the side of the block the player interacted with is has free space above and below it the Rift Entrance will be created adjacent to the spawner and will not rotate. The key will then be consumed.

Using a Rift Key on a Rift Spawner when there is already an open rift will close the rift without consuming the key.

The Rift Entrance can be entered by moving into it, which will send the player to a fixed location in the nether (unsafe teleportation OHAS warning).

The Rift Key and Rift Spawner can be obtained from the creative menu.

### Implementation notes

In the future the rift closing behaviour should be removed.
Rift Entrances should either close when a complete party is through, or when the Rift is over.

The shader settings are a little hard coded currently. This includes animation, which doesn't tie into Minecraft's animated texture mcmeta file (I gather entities don't support it in general?).

Need to add some animation or effect for when the entrance opens/closes.

Obviously all the key crafting and modifiers are missing currently.

I know there is some desire for a multiblock rift spawner, this would be an extension to the Rift Spawner block. The block determines whether a rift can spawn and what position the entrance should take, so that is where changes for multiblock would occur.

Rift opening/closing sounds, rift open ambient?


src/main/java/com/wanderersoftherift/wotr/WanderersOfTheRift.java
```diff
@@ -138,5 +143,15 @@ public static void onClientSetup(FMLClientSetupEvent event) {
         private static void registerScreens(RegisterMenuScreensEvent event) {
             event.register(ModMenuTypes.RUNE_ANVIL_MENU.get(), RuneAnvilScreen::new);
         }
+
+        @SubscribeEvent
+        public static void registerEntityRenderers(EntityRenderersEvent.RegisterRenderers event) {
```
Patrigan, line 148: Some of the other branches introduced a ClientRegistryEvents class. This should potentially move there. See Rune_anvil

src/main/java/com/wanderersoftherift/wotr/WanderersOfTheRift.java
```diff
@@ -138,5 +143,15 @@ public static void onClientSetup(FMLClientSetupEvent event) {
...
+            event.registerEntityRenderer(ModEntityTypes.RIFT_ENTRANCE.get(), RiftEntranceRenderer::new);
+        }
+
+        @SubscribeEvent
+        public static void registerShaderPrograms(RegisterShadersEvent event) {
```
Patrigan, line 153: Some of the other branches introduced a ClientRegistryEvents class. This should potentially move there. See Rune_anvil

src/main/java/com/wanderersoftherift/wotr/block/RiftSpawnerBlock.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+        return Optional.empty();
+    }
+
+    private boolean allowsPortal(Level level, BlockPos pos) {
+        BlockState state = level.getBlockState(pos);
+        return state.isAir() || state.getCollisionShape(level, pos).isEmpty();
+    }
+
+    public record SpawnLocation(BlockPos blockPos, Vec3 position, Direction direction) {}
```
Patrigan, line 112: I personally do not like public classes inside another class. If it is public, it is best to move it to its own file.

src/main/java/com/wanderersoftherift/wotr/block/RiftSpawnerBlock.java
```diff
@@ -0,0 +1,113 @@
...
+        BlockPos adjacentPos = pos.relative(facing);
+        if (allowsPortal(level, adjacentPos) && allowsPortal(level, adjacentPos.above()) && allowsPortal(level, adjacentPos.below())) {
+            return Optional.of(new SpawnLocation(adjacentPos, adjacentPos.below().getBottomCenter().subtract(facing.getUnitVec3().scale(0.475)), facing));
+        }
+        return Optional.empty();
+    }
+
+    private boolean allowsPortal(Level level, BlockPos pos) {
+        BlockState state = level.getBlockState(pos);
+        return state.isAir() || state.getCollisionShape(level, pos).isEmpty();
```
Patrigan, line 109: Wouldn't this allow the rift to spawn on blocks without collision, like an open Fence Gate.

src/main/java/com/wanderersoftherift/wotr/datagen/ModModelProvider.java
```diff
@@ -93,4 +80,14 @@ public void generateRunegemItem(Item item, ItemModelGenerators itemModels) {
         itemModels.itemModelOutput.accept(item, ItemModelUtils.select(new SelectRuneGemShape(), ItemModelUtils.plainModel(modelLocation), list));
     }
 
+    public static PropertyDispatch createOppositeFacingDispatch() {
```
Patrigan, line 83: Did you check if something like this does not already exist for other facingblocks, like wood?

src/main/resources/assets/wotr/shaders/rift_portal.json
```diff
@@ -0,0 +1,97 @@
...
+    {
+      "name": "Sampler1"
+    }
+  ],
+  "uniforms": [
+    {
+      "name": "ModelViewMat",
+      "type": "matrix4x4",
+      "count": 16,
+      "values": [
```
Patrigan, line 22: Overly zealous json formatter?

src/main/resources/assets/wotr/textures/item/rift_key.png
Patrigan, line 1: Check with the artists for a better key, even if just dev texture?

src/main/java/com/wanderersoftherift/wotr/block/RiftSpawnerBlock.java
immortius: Yes, or water, or long grass, or torches. The state certainly needs to be more permissive than air, but this might be too permissive?

It probably can be refined a bit. In the use case of the portal spawning on a wall mostly there shouldn't be things attached to that wall. The use case of a free standing portal is a little different since it is in the middle of the block space.

src/main/java/com/wanderersoftherift/wotr/datagen/ModModelProvider.java
immortius: There is an oppositeDispatch, but it assumes the initial model faces NORTH. In this case the initial model faces UP.

src/main/resources/assets/wotr/shaders/rift_portal.json
immortius: Probably!

immortius: Yeah, the IntelliJ formatter places a value per line. I'm not inclined to fight it. The value is just a default, Minecraft overwrites it with an actual value during use.

src/main/java/com/wanderersoftherift/wotr/WanderersOfTheRift.java
immortius: Done, moved to `com.wanderersoftherift/wotr/init/client/ClientRegistryEvents`

src/main/java/com/wanderersoftherift/wotr/WanderersOfTheRift.java
immortius: Done, moved to `com.wanderersoftherift/wotr/init/client/ClientRegistryEvents`

src/main/java/com/wanderersoftherift/wotr/block/RiftSpawnerBlock.java
immortius: Done, moved to `com.wanderersoftherift.wotr.entity.portal.PortalSpawnLocation`

src/main/resources/assets/wotr/shaders/rift_portal.json
immortius: I've updated it, but will have to see if it sticks or gets changed on accident in the future.

src/main/resources/assets/wotr/textures/item/rift_key.png
immortius: Threw something in

src/main/java/com/wanderersoftherift/wotr/block/RiftSpawnerBlock.java
iwolfking: What is the issue with limiting to air?

immortius: Mostly little things, like having the portal blocked by a thin layer of carpet, or a little bit of long grass. I can just set it to air for now and we can revisit later.

Attack8 (approved):

PR #54, opened 2025-02-26
Feature: Functional Objective system.

Patrigan:
Objective system based on Wrapper implementation.

Wrapper captures all the events an Objective might be interested in. It extracts the ServerLevel and the ObjectiveData. It then forwards the event to the AbstractObjective stored in the ObjectiveData.

By default the AbstractObjective returns false on all events, a boolean used to determine if the objective is dirtied.

This AbstractObjective has an actual child implementation of any of the objective types. In this case the Stealth objective, that simply increases the tracker as mobs are getting killed.


src/main/java/com/dimensiondelvers/dimensiondelvers/rift/objective/ObjectiveEventWrapper.java
```diff
@@ -0,0 +1,22 @@
...
+@EventBusSubscriber(modid = DimensionDelvers.MODID)
+public class ObjectiveEventWrapper {
+
+    @SubscribeEvent
+    public static void onLivingDeathEvent(LivingDeathEvent event) {
+        if(event.getEntity().level() instanceof ServerLevel serverLevel){
+            LevelRiftObjectiveData data = LevelRiftObjectiveData.getFromLevel(serverLevel);
+            if(data != null && data.getObjective() != null){
+                boolean dirty = data.getObjective().onLivingDeath(event, serverLevel, data);
+                if (dirty) data.setDirty();
```
immortius, line 18: I have some very uncomfortable memories regarding if statements without `{}`s.

immortius (commented):
Overall looks good to me. I've left a little comment on a style issue that probably doesn't matter much,

src/main/java/com/dimensiondelvers/dimensiondelvers/rift/objective/ObjectiveEventWrapper.java
Patrigan: Absolutely true.
Will put some braces around it.

immortius (approved):

iwolfking (approved):
Looks good to merge when we're ready.

PR #56, opened 2025-03-04
Implement some debug commands and a BaseCommand

a1qs:
- This PR adds a BaseCommand, which can be used as a basis for any commands we implement in the future, containing a few basic functionalities like permissions, the name of the subcommand, and makes it a bit easier to implement them
- In addition, it adds a DebugCommand, which currently has two subcommands, one to set the world to a devWorld, changing a few gamerules, and another to get all the itemstack components that the player is holding


src/main/java/com/dimensiondelvers/dimensiondelvers/DimensionDelvers.java
```diff
@@ -31,6 +32,7 @@
 import net.neoforged.neoforge.event.entity.player.PlayerEvent;
 import net.neoforged.neoforge.event.server.ServerStartingEvent;
 import org.slf4j.Logger;
+import org.spongepowered.asm.mixin.Debug;
```
immortius, line 35: This looks like an unused import (from a bit of a wild source)

src/main/java/com/dimensiondelvers/dimensiondelvers/commands/BaseCommand.java
```diff
@@ -0,0 +1,72 @@
...
+        return false; // Default is both client & server
+    }
+
+    /**
+     * Allows subclasses to define subcommands or additional arguments.
+     *
+     * @param builder The command builder to which subcommands should be added.
+     */
+    protected abstract void buildCommand(LiteralArgumentBuilder<CommandSourceStack> builder, CommandBuildContext context);
+    protected abstract String getName(); // This defines the subcommand name used after the mod's namespace.
```
immortius, line 56: A pattern I like to use for these sorts of things is to have `name` as a parameter into the constructor for the base type, and the base type hold the name and a final method for it. e.g.

```
public abstract class BaseFoo {

private final String name;

public BaseFoo(String name) {
   this.name = name;
}

public final String getName() {
   return name;
}
```

This achieves a lot of the same result but without subclasses needed to implement the getName() method (or even have it at all, if you don't need it elsewhere).

```
public class Foo extends BaseFoo {

   public Foo() {
      super("foo");
   }
}
```

This would go for clientSideOnly and permissionLevel too. But, just a thought, both approaches achieve what you are after.

src/main/java/com/dimensiondelvers/dimensiondelvers/commands/DebugCommands.java
```diff
@@ -0,0 +1,86 @@
...
+        return 1;
+    }
+
+    private int getItemStackComponents(CommandContext<CommandSourceStack> stack) {
+        ServerPlayer player = stack.getSource().getPlayer();
+
+        if (player != null) {
+            ItemStack heldItem = player.getMainHandItem();
+
+            if(heldItem.isEmpty()) stack.getSource().sendFailure(Component.translatable("command.dimensiondelvers.get_item_stack_components.invalid_item"));
```
immortius, line 69: I realise we don't have a style guide, but I do suggest avoiding if statements without braces, since it can lead to mistakes in the future.

immortius (commented):
I've left a few comments/suggestions. They might be unnecessarily niggly - just let me know, happy to back down on them.

I do appreciate the javadoc, particular for getPermissionLevel.

src/main/java/com/dimensiondelvers/dimensiondelvers/commands/BaseCommand.java
a1qs: I do understand this, however in this case at least, i think the way it is currently is preferred, as currently you would need to register it in a different class, so you would have to define the parameters elsewhere for the name/etc. while having the command itself alongside its functionality seperate from that. I think having all those in one central location is better

src/main/java/com/dimensiondelvers/dimensiondelvers/DimensionDelvers.java
a1qs: oops, dunno how it got in here lol

src/main/java/com/dimensiondelvers/dimensiondelvers/commands/BaseCommand.java
immortius: With what I'm suggesting the registration would not change, and none of this information would have to appear at the point of registration. I'm suggesting it all goes in the constructor of the implementing class. With the example, Foo's constructor has no p, the name is defined in its call to super.

a1qs: ah i see, i misunderstood what you meant lol, yeah that works for me

src/main/java/com/dimensiondelvers/dimensiondelvers/commands/BaseCommand.java
```diff
@@ -0,0 +1,81 @@
...
+        if (isClientSideOnly() && FMLEnvironment.dist != Dist.CLIENT) {
+            return;
+        }
+
+        // Define the base command with permission level
+        LiteralArgumentBuilder<CommandSourceStack> argumentBuilder = Commands.literal(name)
+                .requires(sender -> sender.hasPermission(permissionLevel));
+
+        this.buildCommand(argumentBuilder, context); // Build subcommands
+        dispatcher.register(Commands.literal(DimensionDelvers.MODID).then(argumentBuilder)); // Register under the mod's namespace
```
Patrigan, line 42: So every command needs to be prefixed by DimensionDelvers? This seems a bit excessive and annoying.

I would make it possible to not have that, as I also do not see a reason for it in many cases. Perhaps some sort of config value that can exclude specific commands from needing that?

src/main/java/com/dimensiondelvers/dimensiondelvers/commands/DebugCommands.java
```diff
@@ -0,0 +1,79 @@
...
+        stack.getSource().sendSuccess(() -> Component.translatable(
+                "command.dimensiondelvers.dev_world_set",
+                "Daylight Cycle", "Weather Cycle", "Mob Spawning", "Wandering Trader Spawning",
+                "Vine Growth", "Fire Tick"
+        ), true);
+
+        return 1;
+    }
+
+    private int getItemStackComponents(CommandContext<CommandSourceStack> stack) {
```
Patrigan, line 57: Isn't this just `/data get entity @s SelectedItem.components`. Or am I missing something?

src/main/java/com/dimensiondelvers/dimensiondelvers/commands/BaseCommand.java
a1qs: sure that can be done, but in which cases would you see this as unnecessary? having modcommands not start with the modid is quite tedious, especially in bigger modpacks where a lot of mods add their own commands

src/main/java/com/dimensiondelvers/dimensiondelvers/commands/DebugCommands.java
Patrigan: Alright, I suppose it has a use then. Not entirely sure how often we would need it, but does not hurt to have it, I suppose.

src/main/java/com/dimensiondelvers/dimensiondelvers/commands/BaseCommand.java
Patrigan: I am actually completely in the opposing opinion. Having to search for the modid, especially when they are not always consisten with the modname, is quite tedious. If I want to create a runegem, I would just want `/runegem create [args]`.

But I can see there are strong differing opinions on it, so if necessary I can let this one slide.

immortius (dismissed):
Looks solid to me now, thank you

immortius (approved):

iwolfking (approved):
Looks good to merge when we are ready.

src/main/java/com/wanderersoftherift/wotr/commands/BaseCommand.java
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * Permission Levels:<br>
+     * - `0` → All players<br>
+     * - `1` → Moderator<br>
+     * - `2` → Game Master (Recommended for most mod-related commands)<br>
+     * - `3` → Admin<br>
+     * - `4` → Server Owner<br>
+     * @see <a href="https://minecraft.wiki/w/Permission_level#Java_Edition">Java Edition: Permission Levels</a>
+     */
+    public static class PermissionLevels {
```
Patrigan, line 74: Why not use the permissions in Commands.java
```java
    public static final int LEVEL_ALL = 0;
    public static final int LEVEL_MODERATORS = 1;
    public static final int LEVEL_GAMEMASTERS = 2;
    public static final int LEVEL_ADMINS = 3;
    public static final int LEVEL_OWNERS = 4;```

Patrigan (changes requested):

src/main/java/com/wanderersoftherift/wotr/commands/BaseCommand.java
a1qs: did not see that not gonna lie, yeah this would be better

Patrigan (approved):

PR #60, opened 2025-03-09
feat(gh-actions): Rework nightly builds and added release action

T2ThatGuy:
Notes:
- Nightly will only keep artefacts for up to 14 days, (probably should be changed to 7 later on)
- Draft releases are created upon pushing a tag starting with v ie v1.0.0

To purge all previous tags, *don't if this will delete the releases or not, might have to be manually done*

I can run the below once reviewed and merged, to delete all nightly releases and tags...
powershell (w/ gh cli)
```powershell
foreach ($release in $(gh release list --limit 3000  --json tagName) | ConvertFrom-Json) {
    gh release delete "${release.tagName}" --yes --cleanup-tag
    git tag -d "${release.tagName}"
}
```

*Once done a quick ping to the devs can be made for everyone to run the below command to clean their local repos of the tags*
`git fetch --tags --prune --prune-tags`


McMelonTV (approved):
LGTM

PR #61, opened 2025-03-14
Introduce spotless

immortius:
In the PR I've added spotless to the project, configured it and run spotlessApply to provide an example of the changes it would make.

I've tried to configure it in line with IntelliJ behaviour where possible, including import order and indentation. There are likely areas that still differ.

I also have not turned on some of the options around auto-arranging class elements (e.g. can arrange all static variables before constructors before methods and that sort of thing) - it is tempting to look at doing this though.

Doesn't seem to be an option to introduce braces for if statements, or I've missed it.

As an additional element, I added a github action to run spotlessCheck when a MR is made against main.


.github/workflows/spotless.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up JDK 21
+        uses: actions/setup-java@v2
+        with:
+          java-version: '21'
+          distribution: 'adopt'
+
+      - name: Grant execute permission for gradlew
```
T2ThatGuy, line 25: You don't need to provide execute permissions on gladlew, it should already be done take a look at the [nightly-build.yml](https://github.com/Dimension-Delvers/dimension-delvers-mod/blob/main/.github/workflows/nightly-build.yml) for example... However, maybe this is due to using v4 of the setup-java action instead of v2 as mentioned in the prior comment

.github/workflows/spotless.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up JDK 21
+        uses: actions/setup-java@v2
```
T2ThatGuy, line 20: Can you update this to be v4 instead of v2 so that its inline with the other workflows?

.github/workflows/spotless.yml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          java-version: '21'
+          distribution: 'adopt'
```
T2ThatGuy, line 23: The other workflows are also using `temurin` as the distribution... Not certain if there is a specific reason for it, but might as well keep them all inline

build.gradle
```diff
@@ -1,138 +1,139 @@
...
-                             mod_license            : mod_license,
-                             mod_version            : mod_version,
-                             mod_authors            : mod_authors,
-                             mod_description        : mod_description]
-    inputs.properties replaceProperties
-    expand replaceProperties
-    from "src/main/templates"
-    into "build/generated/sources/modMetadata"
+	var replaceProperties = [minecraft_version      : minecraft_version,
+							minecraft_version_range: minecraft_version_range,
```
T2ThatGuy, line 123: Not sure I agree with _this_ much indentation, is this intentional? Doesn't seem like that is the case elsewhere?

T2ThatGuy (changes requested):
Other than the 4 points of discussion I brought up, I don't have any issues with the PR...

We could go on and on about the styling but for now I am sure its fine, I would suggest an emphasis on it evolving over time too.

.github/workflows/spotless.yml
immortius: All good points. I had been meaning to double check the action - I had hoped since the tutorial I followed was only 2 months old it would be up-to-date, but clear not (suspect AI generated tutorial now). So yes, will bring this in line with nightly-build.yml

build.gradle
immortius: It is currently using tabs instead of spaces for .gradle file indentation... which I guess github displays as more than 4 spaces. I'll switch it across to spaces instead.

src/main/java/com/wanderersoftherift/wotr/WanderersOfTheRift.java
```diff
@@ -1,149 +1,149 @@
...
-        }
-    }
-}
+package com.wanderersoftherift.wotr;
+
+import com.mojang.logging.LogUtils;
+import com.wanderersoftherift.wotr.commands.InventorySnapshotCommands;
+import com.wanderersoftherift.wotr.config.ClientConfig;
+import com.wanderersoftherift.wotr.gui.screen.RuneAnvilScreen;
+import com.wanderersoftherift.wotr.init.*;
```
T2ThatGuy, line 7: Just to bring up another point of discussion, what are we thinking about wildcard imports?

If I remember correctly from a discussion in vc, it was mentioned we shouldn't be using them

T2ThatGuy (commented):
The workflow looks good now 👍

One final thing for discussion, is wildcard imports as mentioned in the comment

src/main/java/com/wanderersoftherift/wotr/WanderersOfTheRift.java
immortius: I don't think replacing wildcard imports is a capability of spotless - it doesn't have access to the type system when it runs, it is only doing format checks and changes. We might need to pair with checkstyle to detect that.

src/main/java/com/wanderersoftherift/wotr/server/inventorySnapshot/containers/ItemStackHandlerContainers.java
```diff
@@ -108,8 +108,10 @@ public List<ItemStack> remove() {
 
         @Override
         public void applyComponents(DataComponentPatch patch) {
-            // Note: there can actually be multile of a non-stacking item within ItemStackHandler slot,
-            // so pull them out one by one and apply the component patch to them before adding them all back in
+            // Note: there can actually be multile of a non-stacking item within
+            // ItemStackHandler slot,
+            // so pull them out one by one and apply the component patch to them before
+            // adding them all back in
```
iwolfking, line 114: I think it looks great for the code itself. I think the only thing I could comment on, is well, the comments. I personally think the comments itself could afford to have a longer line length as I actually find that a bit better to read as it is breaking it in odd spots, but this is a pretty subjective thing.

T2ThatGuy: Personally for me I take issue with it when it starts appearing off screen during reviews, leading to having to scroll the window across in general exploration on github or dealing with wrapping withing PRs... Perhaps that why it could have been brought in here?

Either way its not a deal breaker just a QOL I am way too used to in other non java projects

src/main/java/com/wanderersoftherift/wotr/WanderersOfTheRift.java
T2ThatGuy: Apologies for the delayed reply, I imagine that could be brought in via another PR at a later date but worth noting down somewhere, unless that is easy enough to do now?

Otherwise no further things from me

src/main/java/com/wanderersoftherift/wotr/server/inventorySnapshot/containers/ItemStackHandlerContainers.java
immortius: I've changed the comment length to be 120 now, in line with code. It looks like javadoc comments will cleanly resize to whatever setting we use, but single line comments will be individually adjusted - so for the specific example here it doesn't know the comments should flow together and I probably should have made it a multiline comment instead.

immortius: Maybe we should the comment line length lower, and just need to remember to use multiline comments where sensible so it will flow correctly.

immortius:
I've added checkstyle now, with just a few checks that aren't handled by spotless:
- Wildcard imports
- Naming checks (e.g. check classes start with capital letter, packages are all lower case, etc)
- Ifs without braces
- Case statements with code that fall through

I haven't fixed the reported cases so the errors can be reviewed.
I would not in particular that checkstyle is complaining about some of the CODEC variables being named in uppercase - this is because they are not set to be final and only static final variables are expected to be named in all upper.

Happy to modify as desired.

src/main/java/com/wanderersoftherift/wotr/WanderersOfTheRift.java
immortius: I've added checkstyle now for review.

T2ThatGuy (changes requested):
Can you merge / rebase of main before looking to merge

Also to note, before I can approve the request the github actions should likely be running successfully. Let me know if you need help finding how you can find it from the latest commit on the PR

iwolfking:
> Nothing more to complain / comment on, does anyone else have any comments?
> 
> @iwolfking etc

No, looks good to me, I like the changes to the comments and probably right that we should just adjust to multi-line comments if they end up odd.

PR #62, opened 2025-03-14
Mcmelon/rune anvil

Patrigan:
I am so so sorry.

This branch...


src/main/java/com/wanderersoftherift/wotr/events/RunegemTooltipEvent.java
```diff
@@ -0,0 +1,42 @@
...
+@EventBusSubscriber(modid = WanderersOfTheRift.MODID, value = Dist.CLIENT, bus = EventBusSubscriber.Bus.GAME)
+public class RunegemTooltipEvent {
+    @SubscribeEvent
+    public static void on(RenderTooltipEvent.GatherComponents event) {
+        List<Either<FormattedText, TooltipComponent>> list = event.getTooltipElements();
+        ItemStack stack = event.getItemStack();
+        if (!stack.has(ModDataComponentType.RUNEGEM_DATA)) return;
+
+        RunegemData runegemData = stack.get(ModDataComponentType.RUNEGEM_DATA);
+        if (runegemData == null) return;
```
immortius, line 32: This feels redundant in the presence of the has check, possibly just remove the has check though.

src/main/java/com/wanderersoftherift/wotr/gui/menu/RuneAnvilSimpleClientContainer.java
```diff
@@ -0,0 +1,9 @@
+package com.wanderersoftherift.wotr.gui.menu;
+
+import net.minecraft.world.SimpleContainer;
+
+public class RuneAnvilSimpleClientContainer extends SimpleContainer implements RuneAnvilContainer {
```
immortius, line 5: I would suggest having size a public static final on the RuneAnvil block entity and just creating a simple container with that size, and then removing both this class and the interface (merging the interface with the entity class).

T2ThatGuy (commented):
Apart from what @immortius has brought up, there isn't really anything else more that I have to add.

Would just like to reiterate not moving imports around!

And add the thought of maybe datapacking the rune types / shapes as a possible future thought as there appears to be a lot of locations that need updating / changing if we are going to be adding new types unless we are not going to be changing them and they are set in stone?

src/main/java/com/wanderersoftherift/wotr/WanderersOfTheRift.java
```diff
@@ -1,11 +1,11 @@
 package com.wanderersoftherift.wotr;
 
+import com.mojang.logging.LogUtils;
```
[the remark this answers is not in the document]
Patrigan: It should remain consistent, yeah. Theoretically, I always use Intellij Optimize Imports. I assumed everyone would do this, tbh. And if everyone does, theoretically it stays consistent. But perhaps something as part of the push flow would be better?

Patrigan:
I agree with the datapacking of the shapes. But should we do it as part of this PR still?

src/main/java/com/wanderersoftherift/wotr/gui/menu/RuneAnvilSimpleClientContainer.java
Patrigan: I am not sure this would be much cleaner, as it would be moving a ton of code into the Menu class. Now it is clean as part of the container, where it belongs, I think.

immortius: I was thinking more a Slot thing than the Menu class. Noting that the simple container is just a client side proxy for the real container I'm not sure how much of this logic it needs. But if it works, it works.

Other broader thoughts:
 - Does the runegem anvil need to be a container - does it need to be a permanent store for items? Or could it be like a crafting bench that has only a temporary container while being used?
 - Has NeoForge's `IItemHandler`/`IModifiableItemHandler` approach been considered? This probably could be looked into in future work though. My experience so far suggests the containers approach is better for temporary inventory style work benches, while IItemHandler is better for permanent inventories.

Patrigan: The first point, I understood that people really like the concept of seeing eachother's items in a workbench. That is why it is implemented that way. I questioned Mel the same thing :D
As for the second, I am sure it has not been considered. My experience with it is rather limited. It feels rather complicating for the purpose.

McMelonTV: As for the first point, the original implementation literally did work like that but people wanted it to store stuff lol

src/main/java/com/wanderersoftherift/wotr/WanderersOfTheRift.java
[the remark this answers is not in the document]
immortius: Not really to be resolved in this branch.

src/main/java/com/wanderersoftherift/wotr/client/tooltip/GearSocketTooltipRenderer.java
```diff
@@ -0,0 +1,111 @@
...
+            int scaledY = (int) (y / scaleFactor);
+
+            if (socket.modifier().isPresent()) {
+                RunegemData data = new RunegemData(socket.shape(), null, RunegemTier.RAW);
+                ItemStack fakeStack = new ItemStack(ModItems.RUNEGEM.get());
+                fakeStack.set(ModDataComponentType.RUNEGEM_DATA, data);
+
+                guiGraphics.renderFakeItem(fakeStack, scaledX, scaledY -1);
+            } else {
+                // I dont know why i need to multiply it by 28, but i have to, do i want to know? No
```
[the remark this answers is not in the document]
immortius: Removed scaling double up.

src/main/java/com/wanderersoftherift/wotr/block/blockentity/RuneAnvilBlockEntity.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    protected void setItems(@NotNull NonNullList<ItemStack> nonNullList) {
+        this.items = nonNullList;
+    }
+
+    @Override
+    protected @NotNull AbstractContainerMenu createMenu(int containerId, @NotNull Inventory inventory) {
+        if (this.level == null) {
```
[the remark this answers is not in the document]
immortius: Switched to guava precondition.

src/main/java/com/wanderersoftherift/wotr/gui/menu/RuneAnvilSimpleClientContainer.java
immortius: After discussion, nothing to change here.

radimous (changes requested):
picking up sword from the rune anvil without applying runes will delete your runegems

Apart from this large issue and few nitpicks I've pointed out in separate comments, this PR looks good to me.

robotmonkey1000 (approved):
Things seem good from my end, i'll approve. Would like the changes that @radimous mentioned though.

src/main/java/com/wanderersoftherift/wotr/events/RunegemTooltipEvent.java
```diff
@@ -0,0 +1,48 @@
...
+@EventBusSubscriber(modid = WanderersOfTheRift.MODID, value = Dist.CLIENT, bus = EventBusSubscriber.Bus.GAME)
+public class RunegemTooltipEvent {
+    @SubscribeEvent
+    public static void on(RenderTooltipEvent.GatherComponents event) {
+        List<Either<FormattedText, TooltipComponent>> list = event.getTooltipElements();
+        ItemStack stack = event.getItemStack();
+
+        RunegemData runegemData = stack.get(ModDataComponentType.RUNEGEM_DATA);
+        if (runegemData == null) return;
+        /*for(Pair<HolderSet<Item>, HolderSet<Modifier>> pair : runegemData.modifierLists()) {
```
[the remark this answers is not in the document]
Patrigan: I was planning to rewrite it, but was unsure how. It felt more like some sort of ponder screen was required

radimous (approved):

Patrigan:
This is an example of a how not to handle PRs.

PR #64, opened 2025-03-18
Merge RiftMapper to main to make Pat and everyone happy

Anorak01:
Pat wants stuff on main, so I'm opening a PR for it
stuff is definitely not final and has some debug stuff inside main class, can remove if need be but is currently used to fill up map data for demo use


Attack8 (changes requested):
Blind approvals woooo

Attack8 (approved):

Issue #36, opened 2025-02-16
Rift Key


Attack8:

immortius:
Spike Post: https://discord.com/channels/1328761294085554176/1341343699350847539

PR #67, opened 2025-03-20
Rift Key Forging Proof of Concept

immortius:
Issue #36

### Summary

This pull request introduces a proof of concept Rift Key forging system. A player can create a Rift Key Forge. Existing items can be set to have "essence" of different types and amounts, and by placing these items in the forge a key can be created. The more total essence, the higher the tier of the key, and whichever type of essence is most prevalent will determine the theme of the key. 

This update also changes the appearance of the rift entrance to an oval shape that grows or shrinks as the player approaches or moves away from it.

### Usage instructions

- A key forge can be crafted with a ender pearl atop an anvil. 
- A rift entrance spawner can be crafted with an ender pearl surrounded by stone at a crafting bench.
- Only a few items are currently set to have essence: most food, stone, and stripped logs.

### Implementation notes

 - Essence values are set up using NeoForge's data map system, which is basically tags with addition values. This allows for additional files that add or remove values. In particular it is possible to set up mod-specific files that are only processed if that mod is present, which will be useful when integrating mods in the future.
 - Datagen is supported for specifying essence values.
 - Essence types are just resource locations from the essence value map. May have to change in the future if there is other information to capture against them.
 - Tier costs are currently set up for each additional tier to cost 3x the last and cap at Tier 5. This could be made configurable in the future.
 - The mechanic for determining theme is just using the highest present value. I imagine we might want to support mixes of essence giving specific themes in the future.
 - Tooltips for essence value are currently only shown when using the forge.


src/main/java/com/wanderersoftherift/wotr/init/ModDataMaps.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+@EventBusSubscriber(modid = WanderersOfTheRift.MODID, bus = EventBusSubscriber.Bus.MOD)
+public class ModDataMaps {
+    public static final DataMapType<Item, EssenceValue> ESSENCE_VALUE_DATA = DataMapType.builder(
+                    WanderersOfTheRift.id("essence_value"),
+                    Registries.ITEM,
+                    EssenceValue.CODEC
+            ).synced(EssenceValue.CODEC, true)
+            .build();
+
```
iwolfking, line 23: Would we want to consider having multiple Essence Values associated with an item? I can see that this limits us to a single value at the moment.

src/main/java/com/wanderersoftherift/wotr/gui/menu/KeyForgeMenu.java
```diff
@@ -0,0 +1,190 @@
...
+public class KeyForgeMenu extends AbstractContainerMenu {
+    public final static int INPUT_SLOTS = 4;
+    private final static int OUTPUT_SLOTS = 1;
+    private final static int PLAYER_INVENTORY_SLOTS = 3 * 9;
+    private final static int PLAYER_SLOTS = PLAYER_INVENTORY_SLOTS + 9;
+    private final static int INPUT_SLOTS_X = 31;
+    private final static int INPUT_SLOTS_Y = 33;
+    private final static int INPUT_SLOT_X_OFFSET = 25;
+    private final static int INPUT_SLOT_Y_OFFSET = 25;
+    private static final List<Integer> TIER_COSTS = ImmutableList.of(20, 60, 180, 540, 1620);
```
iwolfking, line 44: I think hard-coding this is fine for now, but it may be good to think of this being configurable in the future.

iwolfking (commented):
I think everything looks good, I think only thing that may be nice to go ahead and implement is multiple Essence Value support, obviously that requires a bit of refactoring here, but I think that will help give the system some more depth. 

I know at the moment that having multiple essence types is not relevant, so could be good reason to go ahead and merge as-is, but I believe at the end of the day it would be good to have a mix of essence values be relevant.

src/main/java/com/wanderersoftherift/wotr/gui/menu/KeyForgeMenu.java
```diff
@@ -0,0 +1,190 @@
...
+
+    private void updateOutput(ResourceLocation type) {
+        int tier = tierPercent.get() / 100;
+        if (tier == 0 && !resultContainer.isEmpty()) {
+            resultContainer.clearContent();
+        } else if (tier > 0 && resultContainer.isEmpty()) {
+            DataComponentPatch patch = buildKeyComponentPatch(tier, type);
+            ItemStack output = ModItems.RIFT_KEY.toStack();
+            output.applyComponents(patch);
+            output.applyComponents(DataComponentPatch.builder().set(ModDataComponentType.RIFT_TIER.get(), tier).build());
```
Patrigan, line 175: Is this not a repeat of previous line?

src/main/java/com/wanderersoftherift/wotr/init/ModDataMaps.java
Patrigan: I agree with this statement, something to work on PostMVP, for sure.

immortius: I've added support for multiple essence values per item.

src/main/java/com/wanderersoftherift/wotr/gui/menu/KeyForgeMenu.java
immortius: It is. Corrected.

Patrigan (approved):

PR #72, opened 2025-03-27
wout/processor_json

cptflavio:
Continuation of adding POIs


cptflavio:
Of the comments on the commit, only point 5 applies. Rest is already done

Patrigan (approved):

PR #76, opened 2025-04-07
wout/processor_json

cptflavio:


cptflavio:
This uses new processor blocks from donoraffe (cstufts) and a small patch from patrigan

PR #80, opened 2025-04-09
Add Ditto Blocks

### LINKED ISSUE #13: Trap Blocks
Trap Blocks and Spring Block

Attack8:
Closes #13


src/main/java/com/wanderersoftherift/wotr/block/DittoBlock.java
```diff
@@ -0,0 +1,169 @@
...
+			return InteractionResult.PASS;
+		}
+	}
+
+	protected void createBlockStateDefinition(StateDefinition.Builder<Block, BlockState> p_273169_) {
+		p_273169_.add(HAS_ITEM);
+	}
+
+	@Nullable
+	public BlockEntity newBlockEntity(BlockPos p_273396_, BlockState p_272674_) {
```
immortius, line 139: Probably can name these parameters.

src/main/java/com/wanderersoftherift/wotr/block/DittoBlock.java
```diff
@@ -0,0 +1,169 @@
...
+		if (!(blockEntity instanceof DittoBlockEntity dittoBlockEntity))
+			return InteractionResult.PASS;
+
+		if (level.isClientSide)
+			return InteractionResult.SUCCESS;
+
+		ItemStack itemstack1 = dittoBlockEntity.getTheItem();
+		if (!itemstack1.isEmpty()) {
+			ItemStack itemstack = getBlock().toStack();
+			if (dittoBlockEntity.isEmpty()) {
```
immortius, line 63: Is this different from the earlier `itemstack1.isEmpty()` check?

src/main/java/com/wanderersoftherift/wotr/block/DittoBlock.java
```diff
@@ -0,0 +1,169 @@
...
+
+	protected InteractionResult useWithoutItem(BlockState state, Level level, BlockPos pos, Player player, BlockHitResult hitResult) {
+		BlockEntity blockEntity = level.getBlockEntity(pos);
+		if (!(blockEntity instanceof DittoBlockEntity dittoBlockEntity))
+			return InteractionResult.PASS;
+
+		if (level.isClientSide)
+			return InteractionResult.SUCCESS;
+
+		ItemStack itemstack1 = dittoBlockEntity.getTheItem();
```
immortius, line 60: Naming probably could be clearer

src/main/java/com/wanderersoftherift/wotr/block/DittoBlock.java
```diff
@@ -0,0 +1,169 @@
...
+			} else {
+				if (itemstack1.getItem() != getBlock().asItem())
+					Containers.dropContents(level, pos, dittoBlockEntity);
+				else
+					return InteractionResult.PASS;
+				dittoBlockEntity.setTheItem(itemstack);
+			}
+
+			level.playSound((Player) null, pos, SoundEvents.DECORATED_POT_INSERT, SoundSource.BLOCKS, 1.0F, 0.5F);
+			if (level instanceof ServerLevel) {
```
immortius, line 74: Could use `if (level instanceof ServerLevel serverLevel)` here and remove the next line.

src/main/java/com/wanderersoftherift/wotr/block/TrapBlock.java
```diff
@@ -0,0 +1,146 @@
...
+
+    public TrapBlock(Properties properties) {
+        super(properties);
+
+        this.registerDefaultState(this.defaultBlockState().setValue(STAGE, 0));
+        this.registerDefaultState(this.defaultBlockState().setValue(DEACTIVATED, false));
+        this.registerDefaultState(this.defaultBlockState().setValue(WAITING_FOR_TICK, false));
+    }
+
+    protected void tick(@NotNull BlockState state, @NotNull ServerLevel level, @NotNull BlockPos pos, @NotNull RandomSource random) {
```
ImplementsLegend, line 41: override not marked with `@Override`

src/main/java/com/wanderersoftherift/wotr/block/SpringBlock.java
```diff
@@ -0,0 +1,61 @@
...
+            0.672F,
+            0.84F,
+            0.98F,
+            1.1F,
+            1.18F,
+            1.27F,
+            1.35F,
+            1.42F,
+            1.5F,
+    };
```
ImplementsLegend, line 30: Values within the array are mutable.

src/main/java/com/wanderersoftherift/wotr/block/SpringBlock.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public void stepOn(@NotNull Level level, @NotNull BlockPos pos, @NotNull BlockState state, @NotNull Entity entity) {
+        launch(entity, state);
+    }
+
+    public void launch(@NotNull Entity entity, @NotNull BlockState state) {
+        entity.resetFallDistance();
+        if (entity.isShiftKeyDown()) return;
+        double delta = DELTA_AT_STRENGTH[state.getValue(STRENGTH) - MIN_STRENGTH];
```
ImplementsLegend, line 46: It might be better to calculate the value instead of loading it from a table.

src/main/java/com/wanderersoftherift/wotr/block/SpringBlock.java
```diff
@@ -0,0 +1,61 @@
...
+import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.world.level.block.state.StateDefinition;
+import net.minecraft.world.level.block.state.properties.IntegerProperty;
+import net.neoforged.neoforge.registries.DeferredBlock;
+import org.jetbrains.annotations.NotNull;
+
+public class SpringBlock extends DittoBlock {
+    public static final int MAX_STRENGTH = 10;
+    public static final int MIN_STRENGTH = 1;
+    public static final IntegerProperty STRENGTH = IntegerProperty.create("strength", MIN_STRENGTH, MAX_STRENGTH);
```
ImplementsLegend, line 18: Why use `MIN_STRENGTH = 1` insted of `0` and `MAX_STRENGTH = 10` instead of `DELTA_AT_STRENGTH.length`?

src/main/java/com/wanderersoftherift/wotr/block/TrapBlock.java
```diff
@@ -0,0 +1,146 @@
...
+            level.setBlockAndUpdate(pos, state.setValue(DEACTIVATED, false).setValue(STAGE, 0).setValue(WAITING_FOR_TICK, false));
+            return;
+        }
+
+        if (state.getValue(STAGE) + 1 > STAGES) {
+            fail(level, pos, state);
+            return;
+        }
+
+        level.setBlockAndUpdate(pos, state.setValue(STAGE, state.getValue(STAGE) + 1).setValue(WAITING_FOR_TICK, false));
```
ImplementsLegend, line 52: The condition is same as `state.getValue(STAGE) >= STAGES` or `state.getValue(STAGE) == STAGES`.
Alternatively `state.getValue(STAGE) + 1` could be extracted into a variable.

src/main/java/com/wanderersoftherift/wotr/block/TrapBlock.java
```diff
@@ -0,0 +1,146 @@
...
+    public boolean canEntityTick(@NotNull BlockPos pos, @NotNull Entity entity) {
+        if (!this.legalEntity(entity)) {
+            return false;
+        }
+        return entity.onGround() && entity.position().y > (double)((float)pos.getY() + 0.6875F);
+    }
+
+    public void fail(@NotNull Level level, @NotNull BlockPos pos, @NotNull BlockState state) {
+        BlockPos[] positions = {pos.west(), pos.north(), pos.east(), pos.south(), pos.above(), pos.below()};
+        for (BlockPos checked_pos : positions) {
```
ImplementsLegend, line 95: Instead of making an array of posicions and iterating over it, you can iterate over `Direction.values()` and calculate the position as `BlockPos checkedPos = pos.relative(direction);`. Also variable name format is nonstandard

src/main/java/com/wanderersoftherift/wotr/block/TrapBlock.java
Attack8: STAGE should never be more than max stages, so == will suffice

src/main/java/com/wanderersoftherift/wotr/block/SpringBlock.java
Attack8: Can you elaborate? I'm not entirely sure what your point is here

src/main/java/com/wanderersoftherift/wotr/block/SpringBlock.java
Attack8: Can you elaborate? I'm not entirely sure how that would be done

src/main/java/com/wanderersoftherift/wotr/block/SpringBlock.java
immortius: If the array doesn't need to be used outside this class, making it private would largely address the concern I think. The "danger" is just any other class can modify the values, potentially accidentally if they grab a reference into another variable - so it isn't really a constant (worst case I've seen is the "constant" for the Up vector being modified accidentally). Or potentially can use DoubleLists.unmodifiable() instead of a raw array.

src/main/java/com/wanderersoftherift/wotr/block/SpringBlock.java
ImplementsLegend: I meant just replacing `DELTA_AT_STRENGTH[state.getValue(STRENGTH) - MIN_STRENGTH]` with some equation like `state.getValue(STRENGTH)*0.1+0.5`. This is simple enough that it should actually be faster than loading it from array.

src/main/java/com/wanderersoftherift/wotr/block/SpringBlock.java
```diff
@@ -28,6 +25,9 @@ public class SpringBlock extends DittoBlock {
             1.42F,
             1.5F,
     };
+    public static final int MAX_STRENGTH = DELTA_AT_STRENGTH.length;
```
ImplementsLegend, line 28: Since `MIN_STRENGTH` was changed to 0, `MAX_STRENGTH` should also be lowered otherwise ArrayIndexOutOfBoundsException may occur. (`MAX_STRENGTH = DELTA_AT_STRENGTH.length - 1`)

src/main/java/com/wanderersoftherift/wotr/block/TrapBlock.java
```diff
@@ -38,13 +41,14 @@ public TrapBlock(Properties properties) {
...
-    protected void tick(@NotNull BlockState state, @NotNull ServerLevel level, @NotNull BlockPos pos, @NotNull RandomSource random) {
+    @Override
+    protected void tick(BlockState state, ServerLevel level, BlockPos pos, RandomSource random) {
         if (state.getValue(DEACTIVATED)) {
             level.setBlockAndUpdate(pos, state.setValue(DEACTIVATED, false).setValue(STAGE, 0).setValue(WAITING_FOR_TICK, false));
             return;
         }
 
-        if (state.getValue(STAGE) + 1 > STAGES) {
+        if (state.getValue(STAGE) == STAGES) {
```
immortius, line 51: I would suggest >= here just in case the value slops past (perhaps because of a code change while the game is saved that reduces the number of stages).

src/main/java/com/wanderersoftherift/wotr/block/blockentity/DittoBlockEntityRenderer.java
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	@Override
+	public void render(@NotNull DittoBlockEntity blockEntity, float partialTick, PoseStack stack, MultiBufferSource bufferSource, int packedLight, int packedOverlay) {
+		DittoBlock dittoBlock = (DittoBlock)blockEntity.getBlockState().getBlock();
+		if (!dittoBlock.shouldRender(blockEntity.getBlockState())) {
+			return;
+		}
+		if ((blockEntity.getTheItem().getItem() instanceof BlockItem) && blockEntity.getTheItem().getItem() != dittoBlock.getBlock().asItem()) {
+			BlockState blockstate = ((BlockItem)blockEntity.getTheItem().getItem()).getBlock().defaultBlockState();
```
ImplementsLegend, line 39: could use `blockEntity.getTheItem().getItem() instanceof BlockItem blockItem` to make a variable, then change the following condition to `blockItem.getItem() != dittoBlock.getBlock().asItem()` and next line to `BlockState blockstate = blockItem.getBlock().defaultBlockState();`

src/main/java/com/wanderersoftherift/wotr/block/blockentity/DittoBlockEntityRenderer.java
```diff
@@ -0,0 +1,61 @@
...
+			if (blockEntity.getLevel() == null) return;
```
ImplementsLegend, line 40: Does this condition need to be here?

immortius (approved):

src/main/java/com/wanderersoftherift/wotr/item/tools/WotRTweaker.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+    @Override
+    public @NotNull InteractionResult useOn(@NotNull UseOnContext context) {
+        BlockPos pos = context.getClickedPos();
+        Level level = context.getLevel();
+        BlockState state = level.getBlockState(pos);
+        Player player = context.getPlayer();
+
+        if (state.getBlock() instanceof TrapBlock) {
+            level.setBlockAndUpdate(pos, ((TrapBlock) state.getBlock()).getTweak());
```
ImplementsLegend, line 27: Can use `state.getBlock() instanceof TrapBlock trapBlock` and simplify the next line.

src/main/java/com/wanderersoftherift/wotr/item/tools/WotRTweaker.java
```diff
@@ -0,0 +1,48 @@
...
+            return InteractionResult.SUCCESS;
+        }
+        if (state.getBlock() instanceof SpringBlock) {
+            int strength = state.getValue(SpringBlock.STRENGTH);
+            strength++;
```
ImplementsLegend, line 32: Just add 1 to the value on initialization instead of incrementing it on next line.

src/main/java/com/wanderersoftherift/wotr/block/DittoBlock.java
```diff
@@ -0,0 +1,164 @@
...
+			if (dittoBlockEntity.isEmpty()) {
+				dittoBlockEntity.setTheItem(itemstack);
+				fillPercent = (float)itemstack.getCount() / (float)itemstack.getMaxStackSize();
+			} else {
+				if (itemstack1.getItem() != getBlock().asItem()) {
+					Containers.dropContents(level, pos, dittoBlockEntity);
+				}
+				dittoBlockEntity.setTheItem(itemstack);
+				fillPercent = (float)itemstack1.getCount() / (float)itemstack1.getMaxStackSize();
+			}
```
ImplementsLegend, line 111: `dittoBlockEntity.setTheItem(itemstack);` is executed in both branches. Calculation of `fillPercent` is also almost identical in both branches.

ImplementsLegend (commented):
There's also more methods missing `@Override` and some files use tabs instead of 4 spaces for indentation.

PR #81, opened 2025-04-12
Pat/cleanup

Patrigan:
Cleans up the events package.

Cleans up Modifier tooltip code.


src/main/java/com/wanderersoftherift/wotr/gui/tooltip/GearImplicitsTooltipEvent.java
```diff
@@ -35,21 +33,17 @@ public static void on(RenderTooltipEvent.GatherComponents event) {
...
-                MutableComponent cmp = Component.literal("+" + roundedValue + " " + modifierInstance.modifier().getRegisteredName()).withStyle(ChatFormatting.AQUA);
-                toAdd.addLast(new ImageTooltipRenderer.ImageComponent(stack, cmp, WanderersOfTheRift.id("textures/tooltip/attribute/damage_attribute.png")));
+            List<TooltipComponent> tooltipComponents = modifierInstance.getTooltipComponent(stack, ChatFormatting.AQUA);
+            toAdd.addAll(tooltipComponents);
         }
 
 
         for (int i = 0; i < toAdd.size(); i++) {
-            list.add(i + 1, Either.right(toAdd.get(i)));
+            list.add(i + 2, Either.right(toAdd.get(i)));
```
immortius, line 46: I wonder if something like

```
list.addAll(2, toAdd.stream().map(Either::right).toList());
```

would work here.

Patrigan: Probably, I do have some concerns on order, but I think that should generally be okay. Will probably also count for the Sockets.

immortius (approved):

src/main/java/com/wanderersoftherift/wotr/gui/tooltip/GearImplicitsTooltipEvent.java
Patrigan: After consideration, we should really combine the tooltips into one class. a1qs had some remarks on it as well.

PR #87, opened 2025-04-15
Integrate rift themes with key forging

immortius:
### Summary

This pull request adds integration between rift key forging and rift themes, by introducing a recipe system that specifies requirements which ingredients must meet in order to produce a theme. These recipes are prioritised, allowing more specific recipes to override more general recipes.

### Usage instructions

Forge a key at a key forge. If life essence accounts for more than half the essence then a Forest themed key will be created; otherwise a Cave themed key will be created.

### Implementation notes

Data gen is supported for defining recipes. Currently only one recipe supported per theme when doing so (as the theme id is used to determine the output file name).

Both essence and item requirements are supported for recipes. This can mean any combination of:

- Min and max quantities of specific essence types (might be used to limit or exclude specific essence types from the ingredients)
- Min and max percentages of specific essence types (probably the most generally useful, for specifying percentage mixes of essence types)
- Anything supported by ItemPredicate for filtering specific items (if, for example, a particular type of item must be present)


immortius:
Note: this is build atop #97

Patrigan (approved):

PR #89, opened 2025-04-16
feat: arachnophobia accessibilty option replaces spiders with turtles

immortius:
Implements model and sound replacement for spiders when the arachnophobia accessibility option is toggled. This includes cave spiders.

I chose turtles because they have a similar shape profile to spiders. I've added some very basic alt textures for the turtles as well.


immortius:
Linked to issue #91

Attack8 (approved):

PR #97, opened 2025-04-18
Inventory snapshot integration with rift entry

immortius:
### Summary

This PR primarily integrate the capture and removal of inventory snapshots with entry and exit from rifts. When a player changes dimension to a rift a inventory snapshot is captured. If the player safely leaves the rift their snapshot is removed, otherwise they presumably die and have their inventory restored.

Additionally this PR does a little bit of cleanup:
- introduces a registry for holding container type handlers
- cleans up the rift portal entrances and exits entity split by introducing a common base entity
- fixes rift exits deserializing as rift entrances
- removed a duplicate dimension change event for players

### Notes

- If you enter a rift from a rift, or exit from a rift into a rift, the existing snapshot is maintained (in case we want to go deeper).
- Haven't actually retested the sophisticated backpacks interop but nothing fundamentally changed.
- Both entry and exit portals share the same renderer now, which might be all that is needed if they can get information on what to render for the inner portal from the entity.


src/main/java/com/wanderersoftherift/wotr/core/rift/RiftLevelManager.java
```diff
@@ -46,7 +46,11 @@
 
 public class RiftLevelManager {
```
Patrigan, line 47: Should this also be final and private constructor?

Patrigan (approved):

src/main/java/com/wanderersoftherift/wotr/core/rift/RiftLevelManager.java
immortius: I would say so.

PR #100, opened 2025-04-23
Adds support for all gear, tools, and weapons to roll random sockets.…

CSTufts:
### Adding Sockets

- the general method I came up with checks the 'tier' of the socketable item 1-6 and rolls that tier or that tier -1 number of slots. I felt that a better expereince than true random based on max tier, cuz  no one would want to roll a diamond sword and only get 1 socket /feels_bad_man
- more than willing to review that and give some a range as opposed to a flat n or n-1 roll. 
- for netherite it reads in what already rolled on the diamond tool and simply adds 1 socket with a max of 6, I confirmed that it does in fact store and remember socketed gems through this process as well.
- this currently ONLY works when you craft something, if someone wants to review what I did and figure out hot to make it work if you cheat something in feel free or let me know how I could accomplish that? 
- we can always just give sockets some commands after MVP as an alternative

### Tags
- I tried to keep things more general sounding when adding stuff into item tiers, happy to relabel any of it if needbe of course. 
- the tiers and what I added into which tier is purely speculative, we can rework as appropriate

### Tools 
- we can simply remove the tools from the tags I added them in for now 
- I can take a swipe at adding json for the runegems after this goes in, for now I would guess that applying a rungem to a tool isn't gonna do anything (or crash), 
- tested it and it actually works when the gem is defined at least, I suspect once the work Darke gets merged it will probably do more, though we obviously need to review what the gems can roll in tools


CSTufts:
Not needed

PR #102, opened 2025-04-23
feat: makes rift loottables drop runegems

Patrigan:
Adds a rift tier check that can be used by loot tables to only drop when the rift tier is in the correct range.


immortius (approved):

src/main/java/com/wanderersoftherift/wotr/loot/predicates/RiftLevelCheck.java
```diff
@@ -0,0 +1,62 @@
...
+import net.minecraft.server.level.ServerLevel;
+import net.minecraft.world.level.storage.loot.LootContext;
+import net.minecraft.world.level.storage.loot.predicates.LootItemCondition;
+import net.minecraft.world.level.storage.loot.predicates.LootItemConditionType;
+
+/**
+ * A LootItemCondition that checks if the current rift level is between the min (inc) and max (inc) values.
+ */
+public record RiftLevelCheck(int minTier, int maxTier) implements LootItemCondition {
+    public static final MapCodec<RiftLevelCheck> CODEC = RecordCodecBuilder.mapCodec(p_338174_ -> p_338174_
```
immortius, line 17: param name to fix here

PR #109, opened 2025-04-25
feat: add some basic difficulty increases. Not tested for balance.

Patrigan:
Just a simple per tier increase of the most basic stats. This has not been tested for balance.

Should certainly be expanded later.

I considered adding it as a modifier, but flat out adjusting the base value seemed the better approach in my head.

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/rift/RiftDifficultyEvents.java
+++ b/src/main/java/com/wanderersoftherift/wotr/rift/RiftDifficultyEvents.java
@@ -7,6 +7,7 @@ import net.minecraft.server.level.ServerLevel;
 import net.minecraft.world.entity.LivingEntity;
 import net.minecraft.world.entity.Mob;
 import net.minecraft.world.entity.ai.attributes.Attribute;
+import net.minecraft.world.entity.ai.attributes.AttributeInstance;
 import net.minecraft.world.entity.ai.attributes.Attributes;
 import net.minecraft.world.entity.player.Player;
 import net.neoforged.bus.api.SubscribeEvent;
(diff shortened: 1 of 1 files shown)


src/main/java/com/wanderersoftherift/wotr/rift/RiftDifficultyEvents.java
```diff
@@ -0,0 +1,57 @@
...
+    private static void applyDifficultyToEntity(LivingEntity livingEntity, ServerLevel serverLevel) {
+        int tier = RiftData.get(serverLevel).getTier();
+        if (livingEntity instanceof Mob mob) {
+            updateAttribute(mob, Attributes.ATTACK_DAMAGE, getDamageMultiplier(tier));
+            updateAttribute(mob, Attributes.MAX_HEALTH, getHealthMultiplier(tier));
+            updateAttribute(mob, Attributes.MOVEMENT_SPEED, getSpeedMultiplier(tier));
+        }
+    }
+
+    private static void updateAttribute(Mob mob, Holder<Attribute> attackDamage, double tier) {
```
Attack8, line 39: Should rename parameter attackDamage to attribute so that it’s more readable

Attack8 (changes requested):

src/main/java/com/wanderersoftherift/wotr/rift/RiftDifficultyEvents.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+        if (event.getLevel() instanceof ServerLevel serverLevel && RiftData.isRift(serverLevel)
+                && event.getEntity() instanceof LivingEntity livingEntity) {
+            applyDifficultyToEntity(livingEntity, serverLevel);
+        }
+    }
+
+    private static void applyDifficultyToEntity(LivingEntity livingEntity, ServerLevel serverLevel) {
+        int tier = RiftData.get(serverLevel).getTier();
+        if (livingEntity instanceof Mob mob) {
```
immortius, line 32: If it is only going to work for mobs, should the mob check be moved into `onEntitySpawned`?

Patrigan: This should eventually change into a more detailed system that detemrines a difference between player summoned and rift spawned. I kept it to mob to prevent oddities before we have something more complex in place.

immortius (approved):

Attack8 (approved):

Issue #115, opened 2025-04-25
[Rift Objectives] Link objective to rift


robotmonkey1000:
Link the current objective to the active rift.
The objective should be able to be completed by the players within the rift.

PR #118, opened 2025-04-25
Objective integration with rifts (Issue #115)

immortius:
## Summary

This PR integrates objectives with rifts.  Specifically:
 - Objectives are now crafted onto keys as part Key Forging.
 - Key Forge recipes for objectives can now be defined
 - When entering a rift an objective will be generated and tied to the rift
 - When a player leaves a rift (or dies in a rift) a rift overview screen is displayed and rewards (if any) are generated and available from that screen.
 - Rift stats are also captured and displayed in the overview screen.

## Details

-  Introduces ObjectiveTypes, which define a type of objective which can then be configured via datapack. When a rift is created the ObjectiveType is used to generate the ongoing objective, generating and random configuration (e.g. the actually requirements from a range specified in the ObjectiveType).
- ObjectiveType is included in RiftConfig so it can be attached to keys
- The Key Forge recipe system has been abstracted to make it easier to add recipes for new facets in the future.
- Objective rewards are generated from loot tables, with a loot table for each completion status currently (Success, Failure and Escaped).
- Broadcasted RiftEvents were introduced for when rifts are created, closed, or a player dies in a rift.
- When a player enters a rift a snapshot of their current stats is created. When they leave a rift this is used to generate a delta to provide their rift stats.

To assist testing the following was also added
- A command for leaving a rift.
- A test objective which is just kill 1 mob.

## Testing instructions

Current recipe is including 5% dark essence (from skulk related items) will craft a key with the stealth objective. Otherwise a key will have the kill objective. Otherwise objective can be configured on a key with `wotr riftKey objective <objective>`.  The `wotr:testkill` objective is a simple objective requiring 1 kill for testing.

When entering with a kill objective the target monster kills will be shown and a tracking bar displayed. If the requisite kills are achieved the bar will be replaced with "Objective complete" text.

On leaving or dying in a rift the result screen should show, with the completion status of Success if the objective was completed, Failed if the player died, or Escaped if the player left without completing the objective. Rewards are provided on success or failure (the failure reward is a named piece of paper).

## Future Work

- Likely will want different objective success loot tables for different objective types
- More stats.
- Support for more complex objectives.
- Support for objectives that trigger leaving the rift on completion.


src/main/java/com/wanderersoftherift/wotr/core/rift/RiftEvents.java
```diff
@@ -22,9 +23,11 @@ public static void onRiftTick(LevelTickEvent.Post event) {
             for (var playerID : data.getPlayers()) {
```
Patrigan, line 23: We should avoid making super long tick events. The player check should probably move into a separate method.

In general, I also feel like this whole tick method should probably also move to something like RiftData. And the event class should call that directly. We will likely get a lot of stuff in this tick method eventually and some children that need to be called and similar.

src/main/java/com/wanderersoftherift/wotr/core/rift/RiftEvents.java
```diff
@@ -22,9 +23,11 @@ public static void onRiftTick(LevelTickEvent.Post event) {
...
                 if (player == null) {
-                    continue; // Player is offline
+                    // Player is offline
+                    continue;
                 }
                 if (player.isDeadOrDying()) {
+                    NeoForge.EVENT_BUS.post(new RiftEvent.PlayerDied(player, serverLevel, data.getConfig()));
                     toRemove.add(playerID);
                 }
                 if (player.level() instanceof ServerLevel pLevel && !RiftData.isRift(pLevel) && pLevel != serverLevel) {
```
Patrigan, line 33: The RiftData check here is obsolete, as we are already returning at the top if it is not a rift.

src/main/java/com/wanderersoftherift/wotr/core/rift/RiftLevelManager.java
```diff
@@ -35,33 +36,86 @@
...
 
     /**
      * @param id
      * @return Whether a level with the given id exists
      */
     public static boolean levelExists(ResourceLocation id) {
         var server = ServerLifecycleHooks.getCurrentServer();
-        var existingRift = server.forgeGetWorldMap().get(ResourceKey.create(Registries.DIMENSION, id));
-        return existingRift != null;
+        var dimension = server.forgeGetWorldMap().get(ResourceKey.create(Registries.DIMENSION, id));
```
Patrigan, line 62: Can the WorldMap contain null for the ID? Else we can just use a contains in the location.

src/main/java/com/wanderersoftherift/wotr/core/rift/RiftLevelManager.java
```diff
@@ -35,33 +36,86 @@
...
+            return serverLevel;
+        }
+        return null;
+    }
+
+    /**
+     * @param player The player to remove from a rift level
+     * @return Whether the player was successfully removed from a rift
+     */
+    public static boolean removePlayerFromRift(ServerPlayer player) {
```
Patrigan, line 84: Should we use this method also in the tick method in RiftEvents?

src/main/java/com/wanderersoftherift/wotr/core/rift/RiftLevelManager.java
```diff
@@ -35,33 +36,86 @@
...
+        ServerLevel respawnDimension = riftLevel.getServer().getLevel(respawnKey);
+        if (respawnDimension == null) {
+            respawnDimension = riftLevel.getServer().overworld();
+        }
+
+        var respawnPos = riftData.getPortalPos().above();
+        player.teleportTo(respawnDimension, respawnPos.getCenter().x(), respawnPos.getY(), respawnPos.getCenter().z(),
+                Set.of(), player.getRespawnAngle(), 0, true);
+        riftData.removePlayer(player.getUUID());
+        if (riftData.getPlayers().isEmpty()) {
```
Patrigan, line 106: I have seen this piece of code a couple times already, perhaps move into its own method to ensure that all tasks are done the same every time?

src/main/java/com/wanderersoftherift/wotr/datagen/ModRiftObjectiveLootTableProvider.java
```diff
@@ -0,0 +1,77 @@
...
+import net.minecraft.world.level.storage.loot.providers.number.ConstantValue;
+import net.minecraft.world.level.storage.loot.providers.number.UniformGenerator;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.function.BiConsumer;
+
+public record ModRiftObjectiveLootTableProvider(HolderLookup.Provider registries) implements LootTableSubProvider {
+
+    public void generate(BiConsumer<ResourceKey<LootTable>, LootTable.Builder> consumer) {
+        consumer.accept(getResourceKey("rift_objective/success"), LootTable.lootTable()
```
Patrigan, line 27: We should also add the runegem geodes to this. Just because we can. That loottable exists elsewhere though.

Might have an issue with checking the rift tier.

src/main/java/com/wanderersoftherift/wotr/item/riftkey/RiftConfig.java
```diff
@@ -20,34 +23,43 @@
...
  */
-public record RiftConfig(int tier, Optional<Holder<RiftTheme>> theme, Optional<Integer> seed) {
+public record RiftConfig(int tier, Optional<Holder<RiftTheme>> theme, Optional<Holder<ObjectiveType>> objective,
+        Optional<Integer> seed) {
 
     public static final Codec<RiftConfig> CODEC = RecordCodecBuilder
             .create(instance -> instance
                     .group(Codec.INT.fieldOf("tier").forGetter(RiftConfig::tier),
                             RiftTheme.CODEC.optionalFieldOf("theme").forGetter(RiftConfig::theme),
+                            RegistryFixedCodec.create(RegistryEvents.OBJECTIVE_REGISTRY)
```
Patrigan, line 38: Should probably move this line to something like Objective.CODEC. That would be more consistent with everything else.

immortius: Done

src/main/java/com/wanderersoftherift/wotr/core/rift/RiftLevelManager.java
immortius: No, because this method is for players leaving the rifts alive - the behaviour doesn't mesh well (break inventory snapshots for one). Naming could be better. I think I can make things cleaner though.

src/main/java/com/wanderersoftherift/wotr/core/rift/RiftLevelManager.java
immortius: This is actually moving this code to its own method :)

src/main/java/com/wanderersoftherift/wotr/datagen/ModRiftObjectiveLootTableProvider.java
immortius: The player will be in the overworld by this point, right. But the rift tier is available, and could be added as a loot context param. And then the rift tier check can try to grab the from the loot context first and level second?

src/main/java/com/wanderersoftherift/wotr/core/rift/RiftEvents.java
immortius: I have split the post tick event handler into a PlayerChangedDimensionEvent handler and a LivingDeathEvent handler. I've also moved more of the logic into RiftLevelManager (I felt that was the better place to encapsulate the logic, with RiftData mostly managing the data of which players are in the rift).

src/main/java/com/wanderersoftherift/wotr/core/rift/RiftEvents.java
immortius: Gone with the refactor

src/main/java/com/wanderersoftherift/wotr/core/rift/RiftLevelManager.java
immortius: Switched to a contains check. Seems to work fine from testing.

src/main/java/com/wanderersoftherift/wotr/core/rift/RiftLevelManager.java
immortius: Renamed to `returnPlayerFromRift`. It is alongside an onPlayerDeath method.

Patrigan (approved):
Seems fine now, might need a second set of eyes.

PR #123, opened 2025-04-28
Small clean ups

immortius:
- Remove example item and block
- Change json eol to lf, as it is generated that way.
- Split creative tabs between wotr and wotr dev items
- Add recipe for rune anvil


src/main/java/com/wanderersoftherift/wotr/init/ModItems.java
```diff
@@ -92,4 +89,12 @@ public static <T extends Block> DeferredItem<BlockItem> registerSimpleBlockItem(
         return simpleBlockItem;
     }
 
+    public static <T extends Block> DeferredItem<BlockItem> registerSimpleDevBlockItem(
```
Patrigan, line 92: Not entirely satisfied with this change, as this piece of code can be used for all blocks.

Patrigan (changes requested):

src/main/java/com/wanderersoftherift/wotr/init/ModItems.java
Patrigan: Nvm, did not see the original method still exists.

Patrigan (approved):

PR #126, opened 2025-04-30
feat: Copy state from all blocks to all blocks.

Patrigan:


immortius (approved):
If I understand correctly, this is a simplification to how the copyState method works to be more generic/better leverage properties to require less specific code, right?

Patrigan:
Correct. Basically exactly that.

PR #128, opened 2025-05-02
Configurable HUD positioning

immortius:
### Summary

This PR introduces a framework for configuring the display of HUD elements (Layers in NeoForge).  Configuration options include:

 - Visbility
 - Positioning relative to an anchor point - one of each screen corner, center of a screen side or the center of the screen.
 - Orientation or rotation, either horizontal or vertical

These can be configured through the usual client-side configuration file, or through a settings menu. Additionally presets can be provided for hud configurations via resource packs.

### Implementation

The key implementation elements are

- HudElementConfig: This class groups configuration for a single HUD element, for use when defining a Config Spec.
- ConfigurableLayer: Implemented by layers that are configurable. This joins the layer to its config and provides width/height information for display in the settings screen. Registered into a Configurable Layer registry for discovery.
- ConfigurableLayerAdapter: Provides support for linking existing, non-configurable layers to the configuration framework, currently through the FixedSizeLayerAdapter.
- HudPreset: A predefined configuration for the hud that can be provided through a resource pack.

When implementing ConfigurableLayer, a layer calculates its width and height and then uses its linked HudElementConfig to determine its top-left position and (optionally) orientation before rendering.

### Areas for improvements

- There is likely room for one or more base ConfigurableLayer implementations to simplify creating new configurable layers. Currently the layers implementing ConfigurableLayer are left to consume the configuration as they see fit.
- The implementation of orientation in Ability Bar, Mana Bar and Effect Bar could almost certainly be improved - a lot of similar code handling with x and y behaviors swapped.
- The major vanilla layers have been adapted, but the adaptation simply assumes they render to a fixed location. This means the mob effect display cannot be set to flow from the left, and health + armor displays are joined because the armor display is offset when there are more than 10 hearts. More sophisticated adapters are possible and may provide better control, but might ultimately need to replace the vanilla rendering.
- The code for reorientating an existing element that has a default vertical orientation is almost certainly broken (not tested because none exist in vanilla minecraft)


src/main/java/com/wanderersoftherift/wotr/gui/layer/ManaBar.java
```diff
@@ -23,32 +27,57 @@
...
-    private static final int BAR_WIDTH = 5;
-    private static final int FILL_WIDTH = 3;
-    private static final int TOP_HEIGHT = 1;
-    private static final int BOTTOM_HEIGHT = 1;
-    private static final int FILL_X_OFFSET = 1;
+    private static final int BAR_THICKNESS = 5;
+    private static final int FILL_THICKNESS = 3;
+    private static final int START_SIZE = 1;
+    private static final int END_SIZE = 1;
+    private static final int FILL_OFFSET = 1;
```
robotmonkey1000, line 49: This feels like quite a few magic numbers here.

At least for texture sizes, is there no way to pull it from the file(s)?

robotmonkey1000 (commented):
Overall, the code seems clean, and no major issues jump out to me. 
I do see quite a few constants in some of the huds which we may want to look into cleaning up. Especially for anything texture related, it may be worth changing how we approach that so that is the texture is modified, we can more easily find where to blit in the file. 

Or at least documenting some of them so if the art team wants to make any changes it isnt a major code hassle. 

As for animation related values in the huds, we should look into adding a slider for animation  speeds so users can slow them down. 

 Seems solid though on my end!

Patrigan (approved):
Knuckles says: APPROVED!

PR #129, opened 2025-05-02
feat: Ability items now drop in rifts.

Patrigan:
Had to move the abilities to DataGen.

We should look for a solution.


src/main/java/com/wanderersoftherift/wotr/datagen/ModAbilityTagsProvider.java
```diff
@@ -0,0 +1,28 @@
...
+public class ModAbilityTagsProvider extends TagsProvider<AbstractAbility> {
+    // Get parameters from the `GatherDataEvent`s.
+    public ModAbilityTagsProvider(PackOutput output, CompletableFuture<HolderLookup.Provider> lookupProvider) {
+        // Second parameter is the registry key we are generating the tags for.
+        super(output, RegistryEvents.ABILITY_REGISTRY, lookupProvider, WanderersOfTheRift.MODID);
+    }
+
+    @Override
+    protected void addTags(HolderLookup.Provider lookupProvider) {
+        tag(ModTags.Abilities.RIFT_DROPS)
```
immortius, line 23: Looks like
```
        tag(ModTags.Abilities.RIFT_DROPS)
                .add(TagEntry.optionalElement(WanderersOfTheRift.id("dash")))
                .add(TagEntry.optionalElement(WanderersOfTheRift.id("fireball")))
                .add(TagEntry.optionalElement(WanderersOfTheRift.id("heal")));
```
works without the abilities needing to be data gen (but do need the ability registry bootstrapped).

immortius: Experimenting locally, by combining this with using the DeferredHolder approach for the recipes and changing the DeferrableRegistryCodec to eliminate existence checks when encoding, it is possible for data gen to reference non-datagenned abilities. It would be nice if we could set the DataComponent to use different codecs when being used in game to being used in data gen though, so we could keep the existence checks when encoding in game.

src/main/java/com/wanderersoftherift/wotr/datagen/ModChestLootTableProvider.java
```diff
@@ -41,6 +45,7 @@ public void generate(BiConsumer<ResourceKey<LootTable>, LootTable.Builder> consu
                                         .setWeight(40)
                                         .apply(SetItemCountFunction.setCount(UniformGenerator.between(1.0F, 2.0F))))
                                 .add(NestedLootTable.lootTableReference(getResourceKey("rift/runegem")).setWeight(20))
+                                .add(NestedLootTable.lootTableReference(getResourceKey("rift/ability")).setWeight(80))
```
CSTufts, line 48: We should probably chat weights of all these things at some point, not holding up this PR with the thought, but we probably want to review all the weights of the drops.

Patrigan:
> Is the deletion of the existing ability json files missing from this PR?

The PR considers the 3 main abilities "moved", because their contents are very much the same. (they are now in the generated folder)

src/main/java/com/wanderersoftherift/wotr/datagen/ModChestLootTableProvider.java
Patrigan: Oh, yeah, I had increased the drop chance to see if they actually dropped. Probably should be lowered to 10 or 5 even.

Attack8 (approved):

PR #132, opened 2025-05-04
Add sockets to gear on craft

CSTufts:
- Modifies how we are adding sockets to gear from ItemStack mixin to Item Mixin. 
- Adds back a more generic socketable tag to support this by adding all the tags Darke added to a parent socketable tag
- Adds stone tools to main hand socketable tools. 
- Adds a min and max socket to implicitly call on itemstack in case we need it later 
- Adds checks to make sure there aren't sockets on the gear before rolling it.
- removes enchanted books and enchanted gear from vanilla loot tables
- adds sockets rolls to socketable items in vanilla loot pools (1-3 in the end, 1-2 everywhere else)
- adds a small chance to drop raw runegem geodes from vanilla loot pools


src/main/java/com/wanderersoftherift/wotr/loot/lootmodifiers/RemoveEnchantedItemsModifierAndAddSockets.java
```diff
@@ -0,0 +1,49 @@
...
+            LootContext context) {
+
+        // Iterate through the loot and remove enchantments from each item
+        for (ItemStack itemStack : generatedLoot) {
+            if (itemStack.isEnchanted()) {
+                EnchantmentHelper.setEnchantments(itemStack, ItemEnchantments.EMPTY); // Remove all enchantments
+            }
+            if (itemStack.is(SOCKETABLE)) {
+                // Add sockets to the item
+                GearSockets.generateForItem(itemStack, context.getLevel(), 2, 4);
```
Patrigan, line 39: 2 and 4 is a bit high. I would keep it to 1-2. Overworld items should not be better than rift items.

If you are up for a challenge, I would make The End roll with 1-3.

src/main/java/com/wanderersoftherift/wotr/mixin/MixinItem.java
```diff
@@ -0,0 +1,20 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(Item.class)
+public abstract class MixinItem {
+
+    @Inject(method = "onCraftedPostProcess", at = @At("TAIL"))
+    public void addCustomDataOnCraftedPostProcess(ItemStack stack, Level level, CallbackInfo ci) {
```
Patrigan, line 16: Does this also cover the autocrafter? That was the reason why I had it on ItemStack.

src/main/java/com/wanderersoftherift/wotr/loot/lootmodifiers/RemoveEnchantedItemsModifierAndAddSockets.java
```diff
@@ -0,0 +1,49 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import static com.wanderersoftherift.wotr.init.ModTags.Items.SOCKETABLE;
+
+public class RemoveEnchantedItemsModifierAndAddSockets extends LootModifier {
+
+    public static final MapCodec<RemoveEnchantedItemsModifierAndAddSockets> CODEC = RecordCodecBuilder.mapCodec(
+            inst -> LootModifier.codecStart(inst).apply(inst, RemoveEnchantedItemsModifierAndAddSockets::new));
+
+    protected RemoveEnchantedItemsModifierAndAddSockets(LootItemCondition[] conditions) {
```
Patrigan, line 23: Not sure if this should both be in 1 LootModifier.

At the very least, I would sokit the DoApply into different submethods (removeEnchants and applySockets).

src/main/java/com/wanderersoftherift/wotr/loot/lootmodifiers/RemoveEnchantedItemsModifierAndAddSockets.java
```diff
@@ -0,0 +1,49 @@
...
+        for (ItemStack itemStack : generatedLoot) {
+            if (itemStack.isEnchanted()) {
+                EnchantmentHelper.setEnchantments(itemStack, ItemEnchantments.EMPTY); // Remove all enchantments
+            }
+            if (itemStack.is(SOCKETABLE)) {
+                // Add sockets to the item
+                GearSockets.generateForItem(itemStack, context.getLevel(), 2, 4);
+            }
+        }
+        return generatedLoot; // Return the modified loot
```
Patrigan, line 42: Perhaps we can also add a check to see if the implicits on the item are unrolled and if so, roll them. This could reduce the  need to have it in MH.

src/main/java/com/wanderersoftherift/wotr/mixin/MixinItem.java
```diff
@@ -0,0 +1,20 @@
...
+        // Add custom data to the ItemStack when it is crafted
+        GearSockets.generateForItem(stack, level, 1, 3);
```
Patrigan, line 18: Perhaps we can also add a check to see if the implicits on the item are unrolled and if so, roll them. This could reduce the  need to have it in MH.

src/main/java/com/wanderersoftherift/wotr/loot/lootmodifiers/RemoveEnchantedItemsModifierAndAddSockets.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public static final MapCodec<RemoveEnchantedItemsModifierAndAddSockets> CODEC = RecordCodecBuilder.mapCodec(
+            inst -> LootModifier.codecStart(inst).apply(inst, RemoveEnchantedItemsModifierAndAddSockets::new));
+
+    protected RemoveEnchantedItemsModifierAndAddSockets(LootItemCondition[] conditions) {
+        super(conditions);
+    }
+
+    @Override
+    protected @NotNull ObjectArrayList<ItemStack> doApply(
```
Patrigan, line 28: We should probably also return early out of this method if it is running on a rift. The loottable manages the sockets then.

Patrigan (changes requested):

src/main/java/com/wanderersoftherift/wotr/mixin/MixinItem.java
CSTufts: I will need to test, maybe it will need to be on both? I already have it not 're-roll' if the sockets already exist.

src/main/java/com/wanderersoftherift/wotr/loot/lootmodifiers/RemoveEnchantedItemsModifierAndAddSockets.java
CSTufts: I will split them out and call them both then, at very least if needbe we can filter them both differently.

src/main/java/com/wanderersoftherift/wotr/loot/lootmodifiers/RemoveEnchantedItemsModifierAndAddSockets.java
CSTufts: would you detect the dim here or filter them out in different glm configs?

src/main/java/com/wanderersoftherift/wotr/loot/lootmodifiers/RemoveEnchantedItemsModifierAndAddSockets.java
CSTufts: I wanted to chat with you on implicits before I work on it and make sure you and I are on the same page so I don't do a bunch of work for naught

src/main/java/com/wanderersoftherift/wotr/loot/lootmodifiers/RemoveEnchantedItemsModifierAndAddSockets.java
Patrigan: personally detect them here.

CSTufts: moved to other class

src/main/java/com/wanderersoftherift/wotr/mixin/MixinItem.java
CSTufts: @Patrigan seems to work with the crafter based on what is in the branch, seems a bit of a stretch to use it to craft tools, but good to know that the mixin seems to support it 😆 
gonna resolve this one

src/main/java/com/wanderersoftherift/wotr/mixin/MixinItem.java
CSTufts: going to handle this after this PR is reviewed and merged

src/main/java/com/wanderersoftherift/wotr/loot/lootmodifiers/RemoveEnchantedItemsModifierAndAddSockets.java
CSTufts: added this, though the different way I am handling loot conditions should render it unnecessary, we should leave it in cases someone does something funky later...

Patrigan (approved):
Looks good now. To be continued as discussed with Donoraffe in #dev-basement.

PR #133, opened 2025-05-04
Pat/spawner update

Patrigan:
Made a new Trial Spawner implementation, that by default removes the line of sight check.

We may want to upgrade this block later to allow for more control. Things like fzzling spawners


ImplementsLegend:
This would probably be a good time to fix the trial spawner sound spam. No one fixed it yet and seems like this block will also be affected.

Patrigan:
https://bugs-legacy.mojang.com/browse/MC-278480?attachmentSortBy=dateTime

immortius:
Not sure whether better to change the TrialSpawner (which is untouched by this MR) to not play its ambient if the player isn't nearby, or to add a generic check when playing a sound (via the PlaySoundEvent or otherwise) to discard sounds if their distance from the listener exceeds their attenuation distance, given the attenuation mode is linear and canStartSilent on the sound is false.

Attack8 (approved):

PR #142, opened 2025-05-10
Pat/secret fun stuff

Patrigan:
Secret Fun Stuff


src/main/java/com/wanderersoftherift/wotr/entity/projectile/SimpleProjectileConfig.java
```diff
@@ -7,14 +7,15 @@
 import net.minecraft.network.codec.StreamCodec;
 import net.minecraft.resources.ResourceLocation;
 
-public record SimpleProjectileConfig(int projectiles, float velocity, boolean gravityAffected, float gravity,
-        int groundPersistTicks, SimpleProjectileConfigRenderConfig renderConfig) {
+public record SimpleProjectileConfig(int projectiles, int pierce, float velocity, boolean gravityAffected,
```
robotmonkey1000, line 10: Minor: Should we possible rename pierce to pierce_count or match the previous terminology of pierce level? "Pierce" to me seems boolean.

robotmonkey1000 (approved):
Looks good to me! Minor semantic issue but can be fixed later, not gonna stop the merge over it.

src/main/java/com/wanderersoftherift/wotr/entity/projectile/SimpleProjectileConfig.java
Patrigan: Good idea, we should do that

Issue #146, opened 2025-05-10
Trial Spawner Plays Noise Excessively


Attack8:
https://github.com/Dimension-Delvers/dimension-delvers-mod/pull/133#issuecomment-2868163801

PR #159, opened 2025-05-11
fix: #146 disable trial spawner ambient sounds

immortius:
This fix simply disables the playback of the trial spawner ambient sounds. We may want to do something with a little more finesse (like limiting the playback range) which is possible also.

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/fix/DisableTrialSpawnerAmbient.java
+++ b/src/main/java/com/wanderersoftherift/wotr/fix/DisableTrialSpawnerAmbient.java
@@ -1,6 +1,7 @@
 package com.wanderersoftherift.wotr.fix;
 
 import com.wanderersoftherift.wotr.WanderersOfTheRift;
+import com.wanderersoftherift.wotr.core.rift.RiftLevelManager;
 import net.minecraft.client.Minecraft;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.sounds.SoundEvents;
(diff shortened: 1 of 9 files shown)


src/main/java/com/wanderersoftherift/wotr/fix/DisableTrialSpawnerAmbient.java
```diff
@@ -0,0 +1,36 @@
+package com.wanderersoftherift.wotr.fix;
```
CSTufts, line 1: ultimately would we want to add a rift check to this? that way normal trial spawners in trial chambers act as expected?
I just feel like if we eventually move away from extending the trial spawner, this is something that we inevitably forget exists and don't come back to clean up

immortius: Fair point, I'll make that change.

src/main/java/com/wanderersoftherift/wotr/fix/DisableTrialSpawnerAmbient.java
```diff
@@ -0,0 +1,38 @@
...
+            .of(SoundEvents.TRIAL_SPAWNER_AMBIENT_OMINOUS.location(), SoundEvents.TRIAL_SPAWNER_AMBIENT.location());
+
+    private DisableTrialSpawnerAmbient() {
+
+    }
+
+    @SubscribeEvent
+    public static void catchPlaySound(PlaySoundEvent event) {
+        if (TRIAL_AMBIENT_SOUNDS.contains(event.getOriginalSound().getLocation())
+                && Minecraft.getInstance().level.dimension().location().getPath().startsWith("rift_")) {
```
Patrigan, line 34: Why not use riftLevelManager.isRift ? Though I suppose that one does not support client levels and sound is probably client.

immortius: Fair call. My caution was because isRift was a method in RiftData (which would be server-side only), but it doesn't actually use RiftData on review. So I've moved it into RiftLevelManager and switched over to using it.

Patrigan (approved):

PR #166, opened 2025-05-11
fix: balanced jump height and movement speed modifiers

Rev219:


Patrigan:
This has been integrated into the modifier rewrite that happened. Closing the PR.

PR #186, opened 2025-05-15
Updating the Key Forge model

Req101:
Updating the key forge to use my new model for it. Will need someone to update the hit box.


src/main/resources/assets/wotr/models/block/key_forge.json
```diff
@@ -1,306 +1,257 @@
 {
-	"parent": "block/block",
```
immortius, line 2: I would recommend adding the parent line to the model (this is what ensures the block appearance in inventory has the isometric angle instead of side-on).

src/main/resources/assets/wotr/models/block/key_forge.json
```diff
@@ -1,306 +1,257 @@
...
-	"credit": "Req",
+	"credit": "Made with Blockbench",
 	"texture_size": [64, 64],
 	"textures": {
-		"0": "wotr:block/key_forge",
-		"particle": "wotr:block/key_forge"
+		"0": "block/Key_Forge_V2_texture",
```
immortius, line 5: These texture lines also need to be changed to match up with the actual texture identifiers.

src/main/resources/assets/wotr/models/block/key_forge.json
```diff
@@ -1,306 +1,257 @@
 {
-	"parent": "block/block",
-	"credit": "Req",
+	"credit": "Made with Blockbench",
```
immortius, line 2: Up to you if you want to leave these crediting blockbench or add your name here.

immortius:
I can have a look at updating the collision boxes.

immortius:
Oddly minecraft is unable to open the texture:
```java.io.IOException: Bad PNG Signature```

Would be interesting to know how you created/saved the PNG.

Req101:
I think the issue is I saved the texture from my discord upload. I'll try uploading the PNG directly from my tablet when I get home.

Patrigan (approved):

PR #191, opened 2025-05-17
fix: Set mob current health after max health is changed

WarrenGore:
fix: Set mob current health after max health is changed during mob spawn.


WarrenGore:
May be other ways of doing this, but this worked.

Patrigan (approved):

Patrigan:
This is probably the best way.

PR #192, opened 2025-05-17
feat: reworked key forge recipes to be more flexible

immortius:
### Summary

This PR updates key forge recipes to use a unified system, such that there are no longer separate recipe types for objectives and rift themes. Instead Key Forge Recipes apply data components to the key, and the type of data component is specified in the recipe. The key forge finds the highest priority usable recipe for each data component to apply.

This change sets up for easier integration with REI or similar, as it reduces the number of recipes and provides a generic way to show the result (apply the output to a blank key).

As part of this change, the elements of RiftConfig are split into separate Data Components which are combined when opening a rift. Backwards compatibility to existing keys using a RiftConfig DataComponent has been maintained but is deprecated.

### DIFF
--- a/src/generated/resources/data/wotr/wotr/key_forge_recipe/cave.json
+++ b/src/generated/resources/data/wotr/wotr/key_forge_recipe/cave.json
@@ -1,0 +1,7 @@
+{
+  "output": {
+    "type": "wotr:rift_theme",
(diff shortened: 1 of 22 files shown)


Patrigan:
Is this system easily expandable to other datacomponents?

PR #200, opened 2025-05-18
fix: Set leaves to persistent in room/POI processors

WarrenGore:
fix: Set leaves to persistent in room/POI processors to prevent unintentional decay/item drops


WarrenGore:
Feel free to reject if I should keep going with the more important attachment/mushroom processors putting things places they cannot be placed. Just pushed this now if we wanted to get this quick fix for the leaves decay part of it in the upcoming fix release.

Patrigan (approved):

PR #211, opened 2025-05-18
Fix/wtf rune anvil

Patrigan:
Rune anvil bug for no duplicate modifiers and no overlapping tooltips.

### DIFF
--- a/config/checkstyle/suppressions.xml
+++ b/config/checkstyle/suppressions.xml
@@ -5,5 +5,6 @@
         "https://checkstyle.org/dtds/suppressions_1_2.dtd">
 
 <suppressions>
-
+    <suppress checks="MethodName" files="com[\\/]wanderersoftherift[\\/]wotr[\\/]mixin[\\/]"/>
+    <suppress checks="MethodName" files="com[\\/]wanderersoftherift[\\/]wotr[\\/]mixinextension[\\/]"/>
 </suppressions>
--- a/src/main/java/com/wanderersoftherift/wotr/gui/menu/RuneAnvilMenu.java
+++ b/src/main/java/com/wanderersoftherift/wotr/gui/menu/RuneAnvilMenu.java
@@ -148,9 +148,11 @@ public class RuneAnvilMenu extends AbstractContainerMenu {
             GearSocket currentSocket = currentSockets.sockets().get(i);
             ItemStack runegem = slot.getItem();
             RunegemData runegemData = runegem.get(WotrDataComponentType.RUNEGEM_DATA);
+            List<GearSocket> existingSockets = new ArrayList<>(newSockets);
+            existingSockets.remove(i);
             if (!runegem.isEmpty() && runegemData != null
-                    && currentSocket.canBeApplied(gear, level, runegemData, newSockets)) {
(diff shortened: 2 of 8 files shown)


src/main/java/com/wanderersoftherift/wotr/mixinextension/WotrGuiGraphics.java
```diff
@@ -0,0 +1,23 @@
...
+import net.minecraft.world.inventory.tooltip.TooltipComponent;
+import net.minecraft.world.item.ItemStack;
+
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.Optional;
+
+public interface WotrGuiGraphics {
+
+    void wotr$RenderTooltipLeft(
```
immortius, line 15: I have a feeling that perhaps the before/after comparison displays shouldn't be tooltips, but this is fine.

config/checkstyle/suppressions.xml
```diff
@@ -5,5 +5,6 @@
         "https://checkstyle.org/dtds/suppressions_1_2.dtd">
 
 <suppressions>
-
+    <suppress checks="MethodName" files="com[\\/]wanderersoftherift[\\/]wotr[\\/]mixin[\\/]"/>
```
immortius, line 8: I assume you are happy with this approach?

src/main/java/com/wanderersoftherift/wotr/gui/menu/RuneAnvilMenu.java
```diff
@@ -148,9 +148,11 @@ public void apply() {
             GearSocket currentSocket = currentSockets.sockets().get(i);
             ItemStack runegem = slot.getItem();
             RunegemData runegemData = runegem.get(WotrDataComponentType.RUNEGEM_DATA);
+            List<GearSocket> existingSockets = new ArrayList<>(newSockets);
```
immortius, line 151: Are there situations where a runegem cannot apply at all because all its possibilities have already been applied? I guess that might be a separate issue.

immortius (approved):

src/main/java/com/wanderersoftherift/wotr/mixinextension/WotrGuiGraphics.java
Patrigan: Yeah, same, just unsure how to show it otherwise atm.

config/checkstyle/suppressions.xml
Patrigan: Yeah. Looks perfect, only location where we allow it is in Mixins.

Though Mojang uses it in some other places too, so sometimes it is a hassle when we copy code :D

src/main/java/com/wanderersoftherift/wotr/gui/menu/RuneAnvilMenu.java
Patrigan: Right now, these situations indeed exist. And they are rather common. Health runegem, for example, can not apply to weapons. Once we have more stuff, like leech and armor penetration, the worst ones in this will be better.

Issue #203, opened 2025-05-18
Item frames in rifts


FireKidy:
Someone decieded to try and use item frames in the entrance room and well they just break

![Image](https://github.com/user-attachments/assets/3fbb541a-2a02-427f-909e-b28bbd86155c)

![Image](https://github.com/user-attachments/assets/1438dde3-a294-4003-9cd9-70937852c1d2)

WarrenGore:
Interesting, I've never seen the one in the tavern room break and that's been there from the first playable rift.

WarrenGore:
Okay, this is just a silly one. The room is set to save entities...there must has been a dropped item frame in the room when it was last saved.

I noticed the item frame was only ever showing at the diorite statue and always in the same place. I reproduced the same thing on the build server, then saved it again. It no longer had an item frame as shown in the bug pictures. Further checked by dropping an item frame on the ground myself, saving the room, and confirming it did spawn in the room on the build server when generated.

Will check to see that this room gets repulled before closing.

FireKidy:
Ahh ok, makes sense

PR #212, opened 2025-05-18
wout/build-update-0.1.2

cptflavio:


WarrenGore:
@cptflavio Can you please repull wotr:rift/room/portal/333/portal-1-grimm before this gets merged? See #203 , it should be good on the build server now.

Issue #201, opened 2025-05-18
Spawners need line of sight


FireKidy:
There's multiple POIs that cannot spawn mobs until a block is broken. 

![Image](https://github.com/user-attachments/assets/8fba6f01-5141-494b-a9e8-9c97b85d5f98)

![Image](https://github.com/user-attachments/assets/555c77b2-fbd9-40f5-a25d-80e1a7bae3b2)

PR #215, opened 2025-05-21
fix: (#201) Make trial spawner behave better when covered

Patrigan:
Several changes to our Rift Mob Spawner.


Patrigan:
To be clear, I opted for a copy of the current trial spawner code and did not yet take the time to clean all the Mojank.

Patrigan:
fixes (#201

src/main/java/com/wanderersoftherift/wotr/block/blockentity/riftmobspawner/RiftMobSpawner.java
```diff
@@ -0,0 +1,491 @@
...
+import net.minecraft.world.phys.shapes.CollisionContext;
+import net.neoforged.neoforge.common.extensions.IOwnedSpawner;
+import net.neoforged.neoforge.event.EventHooks;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+
+public final class RiftMobSpawner implements IOwnedSpawner {
```
immortius, line 62: I assume it wasn't worth/wasn't possible to unfinalize and inherit the TrialSpawner, or wrap one.

I wonder if we should diverge further at this point and remove the ominous behaviour?

src/main/java/com/wanderersoftherift/wotr/block/blockentity/riftmobspawner/RiftMobSpawnerData.java
```diff
@@ -0,0 +1,350 @@
...
+import javax.annotation.Nullable;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+import java.util.function.Function;
+
+public class RiftMobSpawnerData {
```
immortius, line 48: Not sure which parts are changed for these

Patrigan: Had to change the state to our own class and it was hardcoded in this class .

immortius (approved):

Issue #199, opened 2025-05-17
[Bug]: Firetick not dissabled in rifts


DefinitelyNotKira:
### What happened?

Firetick needs to be disabled in rifts. Rifts are exploding.

### Reproduction Steps

Light forest theme up.

### Version

0.1.1

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #216, opened 2025-05-21
fix: (#199) fire tick

Patrigan:


src/main/java/com/wanderersoftherift/wotr/core/rift/RiftLevelManager.java
```diff
@@ -322,11 +325,21 @@ private static ServerLevel createRift(
         }
 
         int seed = config.seed().orElseGet(() -> new Random().nextInt());
+        LevelSettings levelSettings = worldData.getLevelSettings();
+        LevelSettings riftLevelSettings = new LevelSettings(
+                "rift", levelSettings.gameType(), levelSettings.hardcore(), levelSettings.difficulty(),
+                levelSettings.allowCommands(), levelSettings.gameRules(), levelSettings.getDataConfiguration(),
```
immortius, line 331: Would it be possible to customise the game rule for the level at this point by creating a new GameRules, copying settings across and then setting fire tick?

immortius (commented):
This approach does not carry across save and reload.

src/main/java/com/wanderersoftherift/wotr/core/rift/RiftLevelManager.java
immortius: I guess the mixin might be the only way if you need to tweak the rule when loading the level too

Patrigan:
I closed this PR, because it was getting too complex to load our own gamerules for the dimension. The biggest issue is that we can not change the loading of the level without some dirtyness. MinecraftServer.java(L417) needs an exception for our RiftDimensions.

Will switch to a mixin in FireBlock and LavaFluid.

PR #233, opened 2025-05-23
Jyu/rift themes

jyutta:
Chnages to the nether theme and new Skyris theme. Both use modded blocks.


cptflavio (approved):
Seems good to me. Might need a look from a smart person tho, like if something should be added because of mods

jyutta:
Keep this for now, I haven't copied one of the themes anywhere else yet lol.

PR #235, opened 2025-05-24
Desert theme (vanilla)

jyutta:


cptflavio (approved):
Perhaps the key essence requirements are a bit low, when we have another death key. 
Code is fine

jyutta:
Also swamp theme now.

Issue #234, opened 2025-05-23
Crashed in SP entering rift.


CyberDay1:
[crash-2025-05-23_16.32.05-server.txt](https://github.com/user-attachments/files/20418440/crash-2025-05-23_16.32.05-server.txt)

Attached crashlog.

PR #238, opened 2025-05-25
fix: (#237) saving of theme in rift data

### LINKED ISSUE #237: 0.1.3 - Expanded Single Player Crash in Forest Theme
crash and latest log attached

[crash-2025-05-24_20.32.35-server.txt](https://github.com/user-attachments/files/20428716/crash-2025-05-24_20.32.35-server.txt)

[latest.log](https://github.com/user-attachments/files/20428717/latest.log)

top line of the stacktrace:

```
java.lang.IllegalStateException: Do not know how to append a primitive value "wotr:forest" to {BannedPlayers:[],Config:{BannedPlayers:[],Players:["5d8f87a1-f0e9-4aa4-ad94-c0d3cfd10cb6"],PortalDimension:"minecraft:overworld",PortalPos:24649L,objective:"wotr:kill",theme:"wotr:forest",tier:1},Players:["5d8f87a1-f0e9-4aa4-ad94-c0d3cfd10cb6"],PortalDimension:"minecraft:overworld",PortalPos:24649L}

```

immortius:
Closes #237

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/core/rift/RiftData.java
+++ b/src/main/java/com/wanderersoftherift/wotr/core/rift/RiftData.java
@@ -77,15 +77,13 @@ public class RiftData extends SavedData { // TODO: split this
                 .forEach(player -> bannedPlayers.add(UUID.fromString(player.getAsString())));
         RiftConfig config = new RiftConfig(0);
         if (tag.contains("Config")) {
-            config = RiftConfig.CODEC
-                    .parse(registries.createSerializationContext(NbtOps.INSTANCE), tag.getCompound("Config"))
+            config = RiftConfig.CODEC.parse(registries.createSerializationContext(NbtOps.INSTANCE), tag.get("Config"))
                     .resultOrPartial(x -> WanderersOfTheRift.LOGGER.error("Tried to load invalid rift config: '{}'", x))
                     .orElse(new RiftConfig(0));
         }
         Optional<Holder<RiftTheme>> theme = Optional.empty();
         if (tag.contains("Theme")) {
-            theme = RiftTheme.CODEC
-                    .parse(registries.createSerializationContext(NbtOps.INSTANCE), tag.getCompound("Theme"))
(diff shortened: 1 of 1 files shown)


ImplementsLegend:
You really didn't have to change this much to fix that issue. Also #234 is the same issue, so I closed it as duplicate.

CSTufts:
Tested and confirmed that I am not getting the same crash, I can't follow the code well enough to approve tho @Patrigan, if you're around to review.

immortius:
> You really didn't have to change this much to fix that issue. Also #234 is the same issue, so I closed it as duplicate.

Maybe not, I took the opportunity to remove some use of Optional too. But I was finding it difficult to work out how to correctly use a Codec with save/loadAdditional when the codec might not be for a compound element, so if you know the approach please do share :)

ImplementsLegend:
I remember @Patrigan talking about changing the theme **to** Optional so it's weird to immediately revert it without even asking him. And about the codecs, you don't even need to care whether it works with compounds specifically. `Codec.encodeStart` gives you some kind of Tag (in this case StringTag), which you can put into the compound you're saving to. Then `CompoundTag.get` gives you that same Tag which you can just give to the codec.

immortius:
> I remember @Patrigan talking about changing the theme **to** Optional so it's weird to immediately revert it without even asking him. And about the codecs, you don't even need to care whether it works with compounds specifically. `Codec.encodeStart` gives you some kind of Tag (in this case StringTag), which you can put into the compound you're saving to. Then `CompoundTag.get` gives you that same Tag which you can just give to the codec.

That was the missing piece, thank you. I think looked into `Codec.encodeStart` at some point but ran into issues, but was able to get it working on second approach.

On the rest, this is a Merge Request. It is asking for feedback and permission, and for knowledge sharing in both directions. Nothing has been changed until it is approved and merged. Patrigan hasn't been around the last couple of days that I've seen and is in the complete opposite timezone to me (as are most) so I feel it best to drop a MR and let it be reviewed by Patrigan or others when they are available.

PR #243, opened 2025-06-01
Ability bench improvements

immortius:
### Summary

This PR makes two changes to improve the Ability Bench

1. Adds checks for whether an ability still has upgrades to be selected before presenting the option to level up, avoiding situations where a player could purchase an upgrade when there are none remaining.
2. Makes the Skill Thread slot persistent and extends its storage to 256.

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/gui/menu/AbilityBenchMenu.java
+++ b/src/main/java/com/wanderersoftherift/wotr/gui/menu/AbilityBenchMenu.java
@@ -5,13 +5,13 @@ import com.wanderersoftherift.wotr.abilities.attachment.AbilitySlots;
 import com.wanderersoftherift.wotr.abilities.upgrade.AbilityUpgrade;
 import com.wanderersoftherift.wotr.abilities.upgrade.AbilityUpgradePool;
 import com.wanderersoftherift.wotr.block.blockentity.AbilityBenchBlockEntity;
-import com.wanderersoftherift.wotr.block.blockentity.LargeCountItemHandler;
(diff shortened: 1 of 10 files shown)


src/main/java/com/wanderersoftherift/wotr/gui/menu/AbilityBenchMenu.java
```diff
@@ -284,4 +309,73 @@ public boolean stillValid(@NotNull Player player) {
...
+            int startIndex,
+            int endIndex,
+            boolean reverseDirection) {
+        boolean result = false;
+        int i = startIndex;
+        if (reverseDirection) {
+            i = endIndex - 1;
+        }
+
+        if (stack.isStackable()) {
```
Patrigan, line 325: Can we split this method into smaller ones? It feels a bit big.

src/main/java/com/wanderersoftherift/wotr/gui/menu/AbilityBenchMenu.java
```diff
@@ -39,23 +43,28 @@ public class AbilityBenchMenu extends AbstractContainerMenu {
 
     private final ContainerLevelAccess access;
     private final SimpleContainer inputContainer;
+    private final DataSlot canLevel;
```
Patrigan, line 46: DataSlots are really ugly :(

Patrigan: This comment just for complaining, you can safely ignore this.

Patrigan (approved):

PR #250, opened 2025-06-14
feat: add QuickMover api and adapt menus to use it.

immortius:
Extracts the QuickMover api from the guild shop work and improves it:
* Better compatibility with Neoforge's InventoryHandlers
* Prevent stacks both moving into a menu slot and swapping between player inventory and hot bar.

The QuickMover api provides a simpler and more consistent way of defining the menu quickMove function. This should remove a lot of boiler plate and error prone code across menus - particularly given the implementation provided in NeoForge documentation is not actually correctly compatible with InventoryHandlers.

A QuickMover is defined as so:

```
mover = QuickMover.create(this)
                // By specifying the slot where player slots begin (if there are player slots in this menu) automatically adds player inventory/hotbar swapping
                .forPlayerSlots(7)
                // First moving into the first seven slots (menu specific slots)
                .tryMoveTo(0, 7)
                // And then define where the menu specific slots try to quick move to
                .forSlots(0, 7)
                .tryMoveToPlayer()
                .build();
```

and then the quick move method for the menu is just

```
    @Override
    public @NotNull ItemStack quickMoveStack(@NotNull Player player, int index) {
        return mover.quickMove(player, index);
    }
```

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/gui/menu/AbilityBenchMenu.java
+++ b/src/main/java/com/wanderersoftherift/wotr/gui/menu/AbilityBenchMenu.java
@@ -19,7 +19,6 @@ import net.minecraft.world.entity.player.Inventory;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.inventory.AbstractContainerMenu;
 import net.minecraft.world.inventory.ContainerLevelAccess;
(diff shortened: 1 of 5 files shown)


Patrigan (approved):
sonic-boom-knuckles-1

PR #254, opened 2025-06-16
Jyu/non grav blocks

jyutta:


immortius:
I would suggest that rather than including copies of the block textures, should generate models using the existing minecraft textures (this will also be more compatible with texture packs). Which... datagen doesn't make as easy as it should be I don't think.

ImplementsLegend:
Last time I checked sugarcane didn't grow on gravel (but bamboo does for whatever reason)

PR #255, opened 2025-06-17
Added No Gravity Blocks

jyutta:
A new PR, because new branch, I totally broke something in the previous one and couldn't be bothered to figure out what. But hey, we using vanilla textures now and sugarcane can no longer grow on gravel. (Thank you for the comments on the last pr.)

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/datagen/WotrBlockLootTableProvider.java
+++ b/src/main/java/com/wanderersoftherift/wotr/datagen/WotrBlockLootTableProvider.java
@@ -77,7 +77,6 @@ public class WotrBlockLootTableProvider extends BlockLootSubProvider {
         );
     }
 
-
     @Override
     protected @NotNull Iterable<Block> getKnownBlocks() {
         return WotrBlocks.BLOCKS.getEntries().stream().map(Holder::value)::iterator;


src/main/java/com/wanderersoftherift/wotr/datagen/WotrBlockLootTableProvider.java
```diff
@@ -33,6 +34,25 @@ protected void generate() {
         dropSelf(WotrBlocks.MOB_TRAP_BLOCK.get());
         dropSelf(WotrBlocks.SPRING_BLOCK.get());
         add(WotrBlocks.RIFT_MOB_SPAWNER.get(), noDrop());
+        add(WotrBlocks.NOGRAVGRAVEL.get(), createSingleItemTable(Blocks.GRAVEL));
```
Patrigan, line 37: Update the loottable to add the riftcheck.

Patrigan: Requires addition of a proper Rift LootItemCondition check See RiftLevelCheck.

src/main/java/com/wanderersoftherift/wotr/datagen/WotrRecipeProvider.java
```diff
@@ -119,6 +119,206 @@ protected void buildRecipes() {
                 .unlockedBy("has_glass_pane", this.has(Blocks.GLASS_PANE.asItem()))
                 .save(this.output, "wotr:ability_heal");
 
+        ShapedRecipeBuilder.shaped(getter, RecipeCategory.BUILDING_BLOCKS, WotrBlocks.NOGRAVGRAVEL.get(), 8)
```
Patrigan, line 122: Chorus fruit?

src/main/java/com/wanderersoftherift/wotr/init/WotrBlocks.java
```diff
@@ -132,6 +132,114 @@ public class WotrBlocks {
                     .sound(SoundType.WOOD)
                     .noOcclusion()));
 
+    public static final DeferredBlock<Block> NOGRAVGRAVEL = registerBlock("nograv_gravel", () -> new Block(
```
Patrigan, line 135: Create a static void method called registerNoGrav, which takes the idString and the SoundType as parameters and then calls registerBlock like here.

src/main/java/com/wanderersoftherift/wotr/datagen/WotrModelProvider.java
```diff
@@ -96,6 +96,81 @@ protected void registerModels(BlockModelGenerators blockModels, @NotNull ItemMod
                         Variant.variant().with(VariantProperties.MODEL, baseRiftSpawnerModel))
                 .with(createFacingDispatchFromUpModel()));
 
+        blockModels.createTrivialBlock(WotrBlocks.NOGRAVGRAVEL.get(),
+                TexturedModel.CUBE.updateTexture(mapping -> mapping.put(TextureSlot.ALL,
+                        ResourceLocation.fromNamespaceAndPath("minecraft", "block/gravel"))));
```
Patrigan, line 101: use ResourceLocation.withDefaultNamespace instead (only takes the location string, not the "minecraft".)

src/main/java/com/wanderersoftherift/wotr/datagen/WotrBlockLootTableProvider.java
immortius: What do you have in mind Pat - I would have thought NoGravGravel should drop normal gravel regardless of dimension?

Patrigan: If we allow the crafting and building in overworld, it is going to drive me insane. I often redo all my blocks, just break them down and rebuild them. With this I would have to continuously recraft the no gravs.

We do want it to drop as normal blocks in rifts, otherwise there is no way to get the normal blocks from rifts.

immortius: Ah, ok, makes sense.

src/main/java/com/wanderersoftherift/wotr/datagen/WotrBlockLootTableProvider.java
```diff
@@ -34,25 +40,40 @@ protected void generate() {
...
+        addRiftConditionalDrop(WotrBlocks.NOGRAVREDCONCRETEPOWDER.get(), Blocks.RED_CONCRETE_POWDER.asItem());
+        addRiftConditionalDrop(WotrBlocks.NOGRAVBLACKCONCRETEPOWDER.get(), Blocks.BLACK_CONCRETE_POWDER.asItem());
+
+    }
+
+    private void addRiftConditionalDrop(Block block, Item riftDrop) {
+        this.add(block, LootTable.lootTable()
+                .withPool(LootPool.lootPool()
+                        .add(AlternativesEntry.alternatives(
+                                LootItem.lootTableItem(riftDrop).when(RiftLevelCheck.riftTier(1, Integer.MAX_VALUE)),
```
immortius, line 72: Probably should be `0, Integer.MAX_VALUE` as 0 is the minimum riftTier.

immortius: Apparently this would include non-rift, so close for now.

Patrigan (approved):

PR #267, opened 2025-06-22
Prevent Mob Drops in Rifts

### LINKED ISSUE #248: [Feat]: Remove mob drops
### Discord username

Danymaddox

### What do you Suggest?

Mob drops inside Rifts are a pain to handle, plus i dont think there's a need for tons of rotten flesh (maybe other resources could be useful, like bones and such, but i dont think its worth the pain of handling those items tbh)

### Version

0.1.2

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Attack8:
Fixes #248


immortius:
Noting this will also stop players dropping items on death in rifts (which would currently be any obtained loot).

Have you considered using a LootModifier?

immortius:
It occurs to me the `LivingDropsEvent` would also work for this, without needing a mixin.

PR #272, opened 2025-06-24
[Feature]: Jigsaw mixin for Shrine & Objective (can delete)

cptflavio:
Changes the target pool of free/3 POIs to Shrines & Objectives. Partly for me to learn, so probably not optimal and may be removed. Also partly to bring Shrines closer to the horizon.


ImplementsLegend:
Why is the config for it in "ClientConfig.java"?
And that mixin won't work with RiftFaster.

cptflavio:
- Yea was kinda afraid it wouldn't work with RiftFaster, but was learning experience too so I'm not too mad. Also gave me an idea on how it works.
- I wanted the chances to not be hardcoded, but configurable. Thought there could be reasons to have more or less shrines. Copilot recommended this spot, but I have no clue.

cptflavio:
I understand, so should I put it in: java/com/wanderersoftherift/wotr/WanderersOfTheRift.java?
Altho I guess if it doesnt work whenever RiftFaster is implemented, it kinda doesnt matter anyway (hope soon) 🙈

immortius:
I think we'ld want to add a ServerConfig for it - currently other than ClientConfig we have `com.wanderersoftherift.wotr.Config` as a "common" config which is a better fit but not perfect. ServerConfigs don't allow changing when connected to a remote server, which is more correct (the common config wouldn't actually affect rift gen when connected to a remote server).

That said, the questions I would ask is - should this be configurable, or is it something that should be controlled by the objective or rift config/rift modifiers. Also, should this be a mixin, or is there a cleaner way to inject these sort of additions into the rift generation?

cptflavio:
I thought maybe difficulty settings should also controll it, but not sure. Makes sense to have it be controlled by the rift. 
I have no experience in minecraft modding, I thought it had to be a mixin because the jigsaw stuff is all minecraft, and it had to intercept the pool.

That said, if its handled by riftfaster, I guess it all changes.

Also, I believe having shrines/objective on POI locations isnt decided upon yet, but it will be likely to have shrines and objective pools interchange, so this (or alternative) logic should be in. Same logic could apply for downscaling POIs, having 5x5 on 7x7 location (for future).

andydirk:
@cptflavio, did you want this PR to be closed

PR #276, opened 2025-07-05
Fix #220

### LINKED ISSUE #220: Gear (equipable) items apply ability modifies when held in main hand
- Get gear (chestplate) with sockets
- add runegeme with health to sockets (just cuz its obvious when you get hearts)
- move gear into hotbar and put in main hand

you should get the added hearts from the chestpiece, even though it is not 'equiped'

ImplementsLegend:
Fixed #220.
Modified ModifierHelper to accept items in other slots than minecraft EquipmentSlot. For example ability slots are now supported.


src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilitySlots.java
```diff
@@ -20,23 +29,43 @@
...
 
     public static final int ABILITY_BAR_SIZE = 9;
-    public static final Codec<AbilitySlots> CODEC = RecordCodecBuilder.create(instance -> instance.group(
+    public static final Codec<AbilitySlots.Data> CODEC = RecordCodecBuilder.create(instance -> instance.group(
             NonNullList.codecOf(ItemStack.OPTIONAL_CODEC).fieldOf("abilities").forGetter(x -> x.abilities),
             Codec.INT.fieldOf("selected").forGetter(x -> x.selected)
-    ).apply(instance, AbilitySlots::new));
+    ).apply(instance, AbilitySlots.Data::new));
+
+    // maybe this should be a separate class
```
immortius, line 37: Yeah. Could have it in the same package and make the `data()` method package private. Or keep it as a private static inner class and sink it down to the bottom. Either way.

src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilitySlots.java
```diff
@@ -103,9 +132,12 @@ public int getSlots() {
         if (!isItemValid(slot, stack)) {
             return stack;
         }
-        if (abilities.get(slot).isEmpty()) {
+        var oldStack = abilities.get(slot);
```
immortius, line 135: Just noting the previous implementation was taking advantage of the max stack size being 1. I guess conceptually we could have ability items that are one-use consumables that stack though?

src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilitySlots.java
ImplementsLegend: What I was thinking with that comment was that we might want some other Attachments to be aware of AttachmentHolder. If so, we would need to write another attachment serializer but the only difference from this one would be different data codec, different constructor and different data-retrival method. So we could just have one serializer that has these three things as parameters.

src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilitySlots.java
ImplementsLegend: Yeah there isn't much of a reason to limit stack size to 1. Consumable abilities could be quite interesting.

src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilitySlots.java
ImplementsLegend: I have considered that too. For now I made it a record with 3 parameters and placed it in WotrAttachments (probably not the best place for it but I don't know what is).

src/main/java/com/wanderersoftherift/wotr/init/WotrAttachments.java
```diff
@@ -56,12 +64,34 @@ public class WotrAttachments {
...
+            implements IAttachmentSerializer<Tag, A> {
+
+        @Override
+        public A read(IAttachmentHolder iAttachmentHolder, Tag tag, HolderLookup.Provider provider) {
+            return attachmentConstructor.apply(iAttachmentHolder,
+                    codec.decode(provider.createSerializationContext(NbtOps.INSTANCE), tag).getOrThrow().getFirst());
+        }
+
+        @Override
+        public @Nullable Tag write(A abilitySlots, HolderLookup.Provider provider) {
```
immortius, line 91: Should rename `abilitySlots`

src/main/java/com/wanderersoftherift/wotr/modifier/WotrEquipmentSlotFromMC.java
```diff
@@ -8,51 +8,71 @@
...
+                TagKey.create(Registries.ITEM, WanderersOfTheRift.id("socketable_boots_slot"))));
+        slots.add(new WotrEquipmentSlotFromMC(EquipmentSlot.LEGS,
+                TagKey.create(Registries.ITEM, WanderersOfTheRift.id("socketable_leggings_slot"))));
+        slots.add(new WotrEquipmentSlotFromMC(EquipmentSlot.BODY,
+                TagKey.create(Registries.ITEM, WanderersOfTheRift.id("socketable_chestplate_slot"))));
+        slots.add(new WotrEquipmentSlotFromMC(EquipmentSlot.CHEST,
+                TagKey.create(Registries.ITEM, WanderersOfTheRift.id("socketable_chestplate_slot"))));
+        slots.add(new WotrEquipmentSlotFromMC(EquipmentSlot.HEAD,
+                TagKey.create(Registries.ITEM, WanderersOfTheRift.id("socketable_helmet_slot"))));
+        SLOTS = slots;
```
immortius, line 65: Would be good to make these two items immutable.

src/main/java/com/wanderersoftherift/wotr/modifier/WotrEquipmentSlotFromMC.java
```diff
@@ -8,51 +8,71 @@
...
-            Map.entry(EquipmentSlot.BODY,
-                    TagKey.create(Registries.ITEM, WanderersOfTheRift.id("socketable_chestplate_slot"))),
-            Map.entry(EquipmentSlot.CHEST,
-                    TagKey.create(Registries.ITEM, WanderersOfTheRift.id("socketable_chestplate_slot"))),
-            Map.entry(EquipmentSlot.HEAD,
-                    TagKey.create(Registries.ITEM, WanderersOfTheRift.id("socketable_helmet_slot")))
-    ));
+public record WotrEquipmentSlotFromMC(EquipmentSlot minecraftSlot, TagKey<Item> tag) implements WotrEquipmentSlot {
+
+    public static final ArrayList<WotrEquipmentSlotFromMC> SLOTS;
```
immortius, line 22: Program to interface?

src/main/java/com/wanderersoftherift/wotr/abilities/AbilityEvents.java
```diff
@@ -28,6 +31,16 @@
 @EventBusSubscriber(modid = WanderersOfTheRift.MODID, bus = EventBusSubscriber.Bus.GAME)
 public class AbilityEvents {
 
+    @SubscribeEvent
+    public static void registerAbilitySlots(ModifierHelper.IterateEquipmentSlotsEvent event) {
```
immortius, line 35: This isn't really a registration event, since it is called every time equipment slots are being iterated. Perhaps we should have a WotrEquipmentSlot type registry though?

src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilitySlots.java
```diff
@@ -153,7 +154,8 @@ public int getSlotLimit(int slot) {
 
     @Override
     public boolean isItemValid(int slot, ItemStack stack) {
-        return stack.has(WotrDataComponentType.ABILITY);
+        return stack.is(WotrItems.ABILITY_HOLDER);
```
immortius, line 157: It has been proposed that abilities would be crafted through essence infusion - meaning that empty ability holders will acquirable - and this change allows empty ability holders to be slotted. I would also be concerned that this will limit the ability to make variant ability sources (like the consumable ones).

src/main/java/com/wanderersoftherift/wotr/init/WotrAttachments.java
```diff
@@ -56,12 +64,34 @@ public class WotrAttachments {
                             .build());
     public static final Supplier<AttachmentType<AbilitySlots>> ABILITY_SLOTS = ATTACHMENT_TYPES.register(
             "ability_slots",
-            () -> AttachmentType.builder(AbilitySlots::new).serialize(AbilitySlots.CODEC).copyOnDeath().build());
+            () -> AttachmentType.builder((holder) -> new AbilitySlots(holder, null))
+                    .serialize(new AttachmentSerializerFromDataCodec<AbilitySlots.Data, AbilitySlots>(
```
immortius, line 68: To help encapsulate the data record (which I don't really feel should be exposed outside of the Attachment class) could have a static method or constant in AbilitySlots that supplies this as an IAttachmentSerializer. Then both the `data` method and record could be private.

src/main/java/com/wanderersoftherift/wotr/init/WotrAttachments.java
```diff
@@ -56,12 +64,34 @@ public class WotrAttachments {
...
+                    .build());
     public static final Supplier<AttachmentType<AttachedEffectData>> ATTACHED_EFFECTS = ATTACHMENT_TYPES.register(
             "attached_effects",
             () -> AttachmentType.builder(AttachedEffectData::new).serialize(AttachedEffectData.CODEC).build());
     public static final Supplier<AttachmentType<EffectDisplayData>> EFFECT_DISPLAY = ATTACHMENT_TYPES
             .register("effect_display", () -> AttachmentType.builder(() -> new EffectDisplayData()).build());
     public static final Supplier<AttachmentType<ManaData>> MANA = ATTACHMENT_TYPES.register("mana",
             () -> AttachmentType.builder(ManaData::new).serialize(ManaData.CODEC).build());
+
+    public static record AttachmentSerializerFromDataCodec<D, A>(Codec<D> codec,
```
immortius, line 80: Could put this in the `codec` package, perhaps rename that package to `serialization`.

src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilitySlots.java
```diff
@@ -25,6 +29,11 @@
 public class AbilitySlots implements IItemHandlerModifiable {
 
     public static final int ABILITY_BAR_SIZE = 9;
+    public static final TagKey<Item> ABILITY_SLOT_ACCEPTED = new TagKey<>(Registries.ITEM,
```
immortius, line 32: Should be in `WotrTags`.

immortius: I think that is basically it though, will approve after that.

PR #277, opened 2025-07-07
Update: New (modded) themes

cptflavio:
Should be disabled without mods loaded. (key generation still possible)
Also slight organization change:
        // <editor-fold desc="WotR Blocks and Items">
        // Add recipes for the mod's blocks and items


cptflavio:
Just thought that we might wanna add (un-comment) dependencies for the mods?

immortius:
> Just thought that we might wanna add (un-comment) dependencies for the mods?

Keeping the number of runtime dependencies low is nice for keeping load times lower iteratively developing things.  But perhaps should add a comment indicating which dependencies are used for themes.

cptflavio:
Right now the dependencies of the mods used are commented in the dependency list (and version list). I/you can un-comment them.

Block_variants
Biomeswevegone
Chisel
Stoneworks

Those are required for the modded themes, but not for the game to run

cptflavio:
Fixed in #280

PR #279, opened 2025-07-11
Update: room update for 0.1.5

cptflavio:


immortius:
Is this just a dump of the latest rooms from the build server? What review or testing would you like for it?

PR #284, opened 2025-07-14
Fix: Fixed tooltip crash for implicit modifiers

iwolfking:
When a null Style is passed in for the tooltip, defaults to gray instead of crashing.

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/modifier/Modifier.java
+++ b/src/main/java/com/wanderersoftherift/wotr/modifier/Modifier.java
@@ -17,16 +17,15 @@ import net.minecraft.world.item.ItemStack;
 
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.stream.Collectors;
 
 public class Modifier {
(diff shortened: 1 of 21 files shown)


src/main/java/com/wanderersoftherift/wotr/modifier/Modifier.java
```diff
@@ -41,6 +42,10 @@ public List<ModifierTier> getModifierTierList() {
     }
 
     public Style getStyle() {
+        if (style == null) {
+            return Style.EMPTY.withColor(ChatFormatting.GRAY);
```
immortius, line 46: Could also make this the default in the codec for when style isn't provided. Then modifier would never have nulls in it to begin with.

iwolfking: Yeah, we were just talking about this. I'm going to update this later to include the style inside the implicit data files properly and then also hopefully implement the codec fix.

immortius (approved):

Issue #278, opened 2025-07-08
[Feat] Mobskins depend on theme


cptflavio:
I would like mobskins to depend on theme, even better when they have themed hats/gear. 
I think its good to have a clear structure with folders, so adding more is easy. Also would be nice if its datapackable.

For testing, zombie noir skin could be used (in art, asset-requests).

cptflavio:
I got a Noir zombie, and a Fedora item (not ArmorItem, because in 21.5 it will be item anyway, and ArmorItem renders full armor set, rather than only the helmet I have. Also had problems with animations)

I dont know how to edit Trial Spawners depending on theme yet

PR #286, opened 2025-07-17
wout/themed-mobskin

cptflavio:
Adds themed mobs to the rift.


cptflavio:
Changed the default spawner to every theme, rather than poi_theme_default. Because I couldnt get it to work when _default calls for trial_spawner, as that will be before the theme jsons, so spawns default mobs, thus prevent themed mobs.

I hope this is the correct solution for #278

cptflavio:
this should use Codecs too

cptflavio:
Now used codec and such. VariantData is usable for other rift mobs. 

The noir helmet should be replaced with a proper item class, so is just a placeholder for now.

src/main/java/com/wanderersoftherift/wotr/entity/EntityAttributeModEvents.java
```diff
@@ -0,0 +1,19 @@
...
+import com.wanderersoftherift.wotr.init.WotrEntities;
+import net.neoforged.bus.api.SubscribeEvent;
+import net.neoforged.fml.common.EventBusSubscriber;
+import net.neoforged.neoforge.event.entity.EntityAttributeCreationEvent;
+
+import static com.wanderersoftherift.wotr.WanderersOfTheRift.MODID;
+import static net.neoforged.fml.common.EventBusSubscriber.Bus;
+
+@EventBusSubscriber(bus = Bus.MOD, modid = MODID)
+public class EntityAttributeModEvents {
```
immortius, line 13: Perhaps this should be in the `wotr.init` package, as `WotrEntityAttributes`?

src/main/java/com/wanderersoftherift/wotr/entity/mob/RiftZombie.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public String getVariant() {
+        return this.entityData.get(DATA_VARIANT);
+    }
+
+    // Set Attributes from VariantData on spawn(+sets hp to max) / load
+    private void applyVariantStats(boolean isInitialSpawn) {
+        RegistryAccess access = this.level().registryAccess();
+        Optional<Registry<VariantData>> registryOpt = access.lookup(WotrRegistries.Keys.MOB_VARIANTS);
```
immortius, line 35: Could use lookupOrThrow here, if we are expecting the registry to exist.

src/main/java/com/wanderersoftherift/wotr/entity/mob/VariantData.java
```diff
@@ -0,0 +1,47 @@
...
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.entity.ai.attributes.Attribute;
+import net.minecraft.world.entity.ai.attributes.AttributeInstance;
+import net.minecraft.world.entity.ai.attributes.Attributes;
+
+import java.util.Map;
+import java.util.Optional;
+
+public record VariantData(Map<String, Double> attributes) {
```
immortius, line 17: Name is a little generic, could be `MobVariantData`?

src/main/java/com/wanderersoftherift/wotr/entity/mob/RiftZombie.java
```diff
@@ -0,0 +1,75 @@
...
+    public RiftZombie(EntityType<? extends RiftZombie> type, Level level) {
+        super(type, level);
+    }
+
+    public String getVariant() {
+        return this.entityData.get(DATA_VARIANT);
+    }
+
+    // Set Attributes from VariantData on spawn(+sets hp to max) / load
+    private void applyVariantStats(boolean isInitialSpawn) {
```
immortius, line 33: Potentially we should mix this into Mob so it doesn't have to be defined for each type of mob... although I guess they need a custom renderer anyhow. Maybe something for future thought.

src/main/java/com/wanderersoftherift/wotr/entity/mob/RiftZombie.java
cptflavio: I think i defaulted to being optional, but youre right the registry has to exist

src/main/java/com/wanderersoftherift/wotr/entity/mob/RiftZombie.java
cptflavio: I think I understand what you mean, but I wouldnt know how to solve that. For now, ill put it in a TODO.

src/main/java/com/wanderersoftherift/wotr/entity/mob/VariantData.java
cptflavio: Yes!

src/main/java/com/wanderersoftherift/wotr/entity/EntityAttributeModEvents.java
cptflavio: Done! Thanks for checking correct location

Issue #293, opened 2025-07-27
Remove limit on tier number from wotr riftkey command


immortius:
Currently the `wotr riftkey tier X` command only supports tiers 1-7 - this is left over from prior to RiftFaster. The limit on rift tier should be removed.

PR #317, opened 2025-08-04
Fix: Removed command 'riftkey tier' limit

cptflavio:
Removes the limit (7) from the riftkey tier command. Issue #293


cptflavio:
Never considered negative. I'll check what happens, maybe fix it. As it's not gameplay, not sure how bad it actually is

ImplementsLegend:
It just uses the code intended for infinite rifts, it *won't* actually be infinite. Although there will be *some* not-very-intended behavior.

PR #319, opened 2025-08-07
Fix: Server Toast error.

cptflavio:
Server didnt work as it has toast, which is client side only. I did use AI and its not the neatest way and wrong locations, so whomever wants to improve go ahead. I just wanted it to work tonight

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/network/quest/QuestGoalUpdatePayload.java
+++ b/src/main/java/com/wanderersoftherift/wotr/network/quest/QuestGoalUpdatePayload.java
@@ -11,6 +11,8 @@ import net.minecraft.network.codec.ByteBufCodecs;
 import net.minecraft.network.codec.StreamCodec;
 import net.minecraft.network.protocol.common.custom.CustomPacketPayload;
 import net.minecraft.resources.ResourceLocation;
+import net.neoforged.api.distmarker.Dist;
+import net.neoforged.api.distmarker.OnlyIn;
 import net.neoforged.neoforge.network.handling.IPayloadContext;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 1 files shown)


ImplementsLegend:
More specifically, the problem was that client-side class was referenced in code that was loaded on the server, although it would never actually run. The part that helped was moving the toast display to a separate class. The rest doesn't help anything.

immortius:
Ah, I see. Do you think it would be cleanest to have a ToastHelper that is called by the payload handler?

cptflavio:
Ah yes, wotrpayloadhandlers didn't change. I tried putting them in if(something.client), but changed back after it didn't work. As I said, was just a quick fix to get it to work on the build server

immortius:
With a view for a pattern that can be applied to avoid this in the future, while still keeping the payload handling code with the payload, my suggestion is something like this: https://github.com/Wanderers-Of-The-Rift/wotr-mod/commit/1d2bcf99200310503be73a9b26ab59e7977aae6f#diff-851a075efc7230ff746323ea51ccf73c727746e3f0b32a563ee9ef19141ef281

Basically have a private static class in the payload to wrap the handling code. This seems to be enough to prevent the server being exposed to the client code.

src/main/java/com/wanderersoftherift/wotr/network/quest/QuestGoalUpdatePayload.java
```diff
@@ -40,15 +40,29 @@ public record QuestGoalUpdatePayload(UUID quest, int goalIndex, int progress) im
...
-                    Minecraft.getInstance().getToastManager().addToast(new QuestCompleteToast(state.getOrigin()));
-                } else if (state.isGoalComplete(goalIndex)) {
-                    Minecraft.getInstance().getToastManager().addToast(new QuestGoalCompleteToast(state, goalIndex));
-                }
-            }
-        });
+        Handler.handle(this, context);
+    }
+
+    // Handler class to shield the server from client code
```
ImplementsLegend, line 46: I don't think this comment answers the question of "why is it separated?" well enough. I think it should include more detailed explanation of this issue, which led to the separation.

immortius: Feel free to replace the comment with one to your liking.

immortius (approved):

Issue #327, opened 2025-08-09
If a player dies in a non-rift dimension they are shown the post-rift reward screen


immortius:
Originally when a player died, they would only be shown the post-rift reward screen if they died in a rift. Currently it is showing for any death.

PR #328, opened 2025-08-09
fix: #327 isEmpty() method for RiftEntryState to correct bad ==s.

immortius:
Assumption that the singleton instance of RiftEntryState would be maintained seems to be bad, I assume the data attachment system serializes/deserializes them into new instances, probably across the new entity creation as a player respawns. Added an isEmpty() method to do the check instead.


ImplementsLegend (approved):
I totally forgot the respawn will create new copy of RiftEntryState even for EMPTY instance when I wrote that code...

PR #336, opened 2025-08-23
Passive/Trigger-based abilities

ImplementsLegend:
Added `TrackedAbilityTrigger` and trigger registry. Currently 3 implementations:
- `DealDamageTrigger`: ability triggers when dealing damage (crude version of ability weapons)
- `TakeDamageTrigger`: ability triggers when taking damage
- `TickTrigger`: ability triggers every tick

Added `AbilityTracker` attachment: Tracks which abilities are available for every type of trigger. Can activate abilities if given an instance of `TrackedAbilityTrigger`. Abilities can be registered from anywhere. Doesn't actually save anything (Abilities should be re-registered when crossing dimensions/on relog).
Added `AbilityModifier` gear modifier. Can be used to add Trigger-based ability to equipment. Consists of ability and a trigger type.
Added `VirtualEquipmentSlot` for gear modifiers that have no actual item. Used by modifiers of attached effects .


src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilityTracker.java
```diff
@@ -0,0 +1,123 @@
...
+
+    public AbilityTracker(IAttachmentHolder holder, @Nullable Data data) {
+        this.holder = holder;
+    }
+
+    public static AbilityTracker forEntity(Entity entity) {
+        return entity.getData(WotrAttachments.ABILITY_TRACKER.get());
+    }
+
+    public static AbilityTracker forEntityNullable(Entity entity) {
```
immortius, line 45: There is a `getExistingData()` method on IAttachmentHolder that achieves the same effect.

src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilityTracker.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+        return entity.getData(type);
+    }
+
+    public static IAttachmentSerializer<Tag, AbilityTracker> getSerializer() {
+        return SERIALIZER;
+    }
+
+    private Data data() {
+        return null;
```
immortius, line 58: Think you missed this :)

src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilityTracker.java
```diff
@@ -0,0 +1,123 @@
...
+        unregisterAbility(abilityModifier.trigger(), abilityModifier.providedAbility(), slot);
+    }
+
+    public void unregisterAbility(
+            Holder<TrackedAbilityTrigger.Type<?>> trigger,
+            Holder<Ability> abilityHolder,
+            @Nullable WotrEquipmentSlot slot) {
+        abilities.get(trigger)
+                .removeIf(trackedAbility -> trackedAbility.ability.equals(abilityHolder)
+                        && slot.equals(trackedAbility.slot));
```
immortius, line 114: `Objects.equal(slot, trackedAbility.slot)` would help handle the null case.

src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilityTracker.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    public void registerAbility(AbilityModifier abilityModifier, @Nullable WotrEquipmentSlot slot) {
+        registerAbility(abilityModifier.trigger(), abilityModifier.providedAbility(), slot);
+    }
+
+    public void registerAbility(
+            Holder<TrackedAbilityTrigger.Type<?>> trigger,
+            Holder<Ability> abilityHolder,
+            WotrEquipmentSlot slot) {
```
immortius, line 94: Have you considered using ModifierSource here rather than slot? That should remove the need for the virtual slot and would be a more natural fit for the extensibility that ModiferSource offers.

src/main/java/com/wanderersoftherift/wotr/abilities/TrackedAbilityTrigger.java
```diff
@@ -0,0 +1,17 @@
...
+import net.neoforged.neoforge.attachment.AttachmentType;
+
+import javax.annotation.Nullable;
+import java.util.function.Supplier;
+
+public interface TrackedAbilityTrigger {
+
+    Type type();
+
+    record Type<T extends TrackedAbilityTrigger>(MapCodec<T> codec,
```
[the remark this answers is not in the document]
immortius: Side question - do these actually need to be serializable? I see at the moment that they are consumed immediately and that none of the extra fields like source are used - not sure what your future intent is. I guess perhaps to track information for the lifespan of the ability for uses cases like an activated on damage ability that retaliates to the source of the damage? Although maybe abilities just need the ability to have an initial target specified that isn't the caster. Not sure.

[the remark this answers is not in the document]
immortius: As a broader reflection, I think there are two distinct feature both in this file that would benefit from being pulled apart more. The first is the configuration of the trigger - this is being handled just by a reference to the Type here, but I imagine there would be further configuration that would be good to be able to define - like the type of damage or amount of damage needed to trigger the ability (e.g. maybe only fire, or maybe only damage > 3, or maybe only if the damage would otherwise kill the entity).

The second is the trigger instance information that is needed to check the trigger.

Or perhaps these are all things you're already planning for future work.

src/main/java/com/wanderersoftherift/wotr/item/ability/AbilityModifier.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public void disableModifier(double roll, Entity entity, ModifierSource source) {
+        AbilityTracker.forEntity(entity).unregisterAbility(this, source.slot());
+    }
+
+    @Override
+    public void applyModifier() {
+        // what is this unused?
```
immortius, line 57: Yeah, we probably can remove this across the board.

src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilityTracker.java
ImplementsLegend: This null is intentional - there are no data to be saved.

src/main/java/com/wanderersoftherift/wotr/abilities/TrackedAbilityTrigger.java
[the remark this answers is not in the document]
ImplementsLegend: One potential reason for the trigger to be serializable is that it could be added to AbilityContext. Haven't thought much more about it. I have also considered adding some kind of trigger predicates, but for now those are just thoughts.

src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilityTracker.java
ImplementsLegend: That `@Nullable` won't be needed if we keep the virtual slots. Or if we replace it with ModifierSource.

ImplementsLegend:
I tried changing things to ~ModifierSource~ **AbilitySource** instead of slots. It was quite a lot of effort...

ImplementsLegend:
I don't think that's nearly enough to make it work properly.

immortius:
What is the problem you are trying to solve?

ImplementsLegend:
I want these abilities to work properly with all current and future features of the ability system. `AbilityStates` and `AbilityCooldowns` required ability to have WotrEquipmentSlot and other parts may require it as well, so either trigger-based abilities will require equipment slot which means that all abilities and modifiers come frome some equipment slot, even if it's not "real", or the rest of ability system stops being so reliant on equipment slots.

src/main/java/com/wanderersoftherift/wotr/init/WotrAttachments.java
```diff
@@ -83,6 +87,12 @@ public class WotrAttachments {
             "ability_states",
             () -> AttachmentType.builder(AbilityStates::new).serialize(AbilityStates.getSerializer()).build());
 
+    public static final Supplier<AttachmentType<? extends AbilityTracker>> ABILITY_TRACKER = ATTACHMENT_TYPES.register(
+            "ability_tracker",
+            () -> AttachmentType.builder(AbilityTracker::new)
+                    .serialize(AbilityTracker.getSerializer())
```
immortius, line 93: I think you could probably just drop the serialize and copy on death, given there is nothing to serialize and nothing to copy.

src/main/java/com/wanderersoftherift/wotr/abilities/sources/ModifierAbilitySource.java
```diff
@@ -0,0 +1,66 @@
...
+    public Holder<Ability> getAbility(Entity entity) {
+        if (base.getModifierEffects(entity).get(effectIndex) instanceof AbilityModifier ability) {
+            return ability.providedAbility();
+        }
+        return null;
+    }
+
+    @Override
+    public AbilityUpgradePool upgrades(Entity entity) {
+        return AbilityUpgradePool.EMPTY; // todo how do we handle upgrades for these abilities
```
immortius, line 46: I guess there's the question of whether they need them. For gears it can just be a pool on the gear. For abilities that give abilities it could pull from the item of origin/ability context, or just not have them.

immortius (approved):
I think this sets us up well for the future, cheers.

PR #343, opened 2025-08-29
(#298) Quest Giver support

### LINKED ISSUE #298: NPC Quest Givers
Add support for NPCs that act as quest givers - open menus when "spoken to" similar to the current Quest Hub block.

immortius:
This PR introduce a MobInteraction attachment, which provides an extension point for altering Mob behavior when the player interacts. This is used to allow any mob to be made a quest giver.

Additionally introduces ValidatingLevelAccess as an extension off of ContainerLevelAccess, moving the stillValid check for menus into this type - this allows the same menu to be opened by different sources and check correctly if the menu should remain open (e.g. if the QuestHub block is still present and near the player, or if the quest giver Mob is alive and nearby)

Testing

Added a `wotr debug makeQuestGiver <entityId>` command. Can be tested by looking an mob and using tab completion to get its id.

Closes #298

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/core/quest/QuestMenuHelper.java
+++ b/src/main/java/com/wanderersoftherift/wotr/core/quest/QuestMenuHelper.java
@@ -1,0 +1,107 @@
+package com.wanderersoftherift.wotr.core.quest;
+
+import com.wanderersoftherift.wotr.gui.menu.ValidatingLevelAccess;
+import com.wanderersoftherift.wotr.gui.menu.quest.QuestCompletionMenu;
+import com.wanderersoftherift.wotr.gui.menu.quest.QuestGiverMenu;
+import com.wanderersoftherift.wotr.init.WotrAttachments;
+import com.wanderersoftherift.wotr.init.WotrRegistries;
+import com.wanderersoftherift.wotr.network.quest.AvailableQuestsPayload;
+import net.minecraft.core.BlockPos;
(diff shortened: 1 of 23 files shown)


ImplementsLegend:
I expected that there would be something like `QuestGiverAttachment` that could contain some additional data like quest pool. Then interacting with anything that has that attachment would open the quest menu. Then, if it has multiple "interactive" attachments, a menu could open to allow selecting specific one.

immortius:
Nothing preventing having a MultiMenu mob interaction that holds multiple others in the future, once the need arises.

The rest - yes, in the future I envision it would be extended to allow configuration but I'ld say that is scope creep from the current task, and ties in to the whole space of guild quests, Small steps :)

src/main/java/com/wanderersoftherift/wotr/core/quest/QuestMenuHelper.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Common handling of opening quest related menus - at least while we have both the quest hub block and npc quest
+ * givers.
+ */
+public final class QuestMenuHelper {
+
+    private static final int QUEST_SELECTION_SIZE = 5;
```
ImplementsLegend, line 32: Maybe in future individual quest givers could define their own selection size.

ImplementsLegend (approved):

PR #350, opened 2025-08-31
Jyu/removal of nogravitems

jyutta:
Removed no-gravity items from everywhere (except themes).

Changed the color theme that uses no-gravity blocks to use the MotR version, and added dependencies so things won’t break when MotR is not present.


build.gradle
```diff
@@ -168,6 +168,9 @@ dependencies {
     runtimeOnly "curse.maven:cryonic-config-1208896:${cryonic_config_version}"
     runtimeOnly "curse.maven:puzzles-lib-495476:${puzzles_lib_version}"
 
+    // Local MOTR dependency (since it’s not published yet)
+    runtimeOnly files("libs/motr-0.0.2.jar")
```
immortius, line 172: This is the one thing that gives me pause. Technically can always just drop motr into the run/mods directory, don't need to have it as a runtime dependency.

Cjzech: If it is not technically needed then are we adding a potential failure point for no reason?

Issue #356, opened 2025-09-10
Split Main Hand Weapons And Tools


WarrenGore:
Currently socketable_main_hand_slot covers both weapons and tools. Split these into distinct tags to allow modifier lists/runegems to apply differently to both tools and weapons.

PR #363, opened 2025-09-17
Add new tags for tools, weapons and hybrid tools/weapons (#356)

Rev219:
Added some new tags that splits the current weapons and tools into 3 categories:

**Weapons** - Items that currently can only be used as weapons, and have limited utility use.

**Tools** - Items that are inefficient to fight with, but are used to do other things, such as mining, digging, etc.

**Hybrids** Items that can both be used to fight and to destroy blocks. Currently only applies to axes.


ImplementsLegend:
The `socketable_main_hand_slot` was also used for determining whether the modifiers should be applied, so removing it completely will break stuff. https://github.com/Wanderers-Of-The-Rift/wotr-mod/blob/a42dbc3ca14efc73769df8e2675c3fafad0c0386/src/main/java/com/wanderersoftherift/wotr/core/inventory/slot/WotrEquipmentSlotFromMC.java#L78
Btw would you mind linking those to the `WotrTags` constants?

ImplementsLegend:
Yeah, having the hybrid as a 3rd type seems excessive... So maybe you could have just the tag and you add that tag to both weapons and tools?

WarrenGore:
Yeah that's what I was thinking. Though it does then beg the question - should swords and axes have the same modifier values? If not (since axe is slower attack speed), will we eventually end up with axe specific values to balance them?

Should bows/crossbows/maces/tridents (which by the way maces and tridents are missing from the weapon tag in this PR but are in main_hand_slot on main) eventually have distinct modifiers from swords?

I think the best step at this stage is complete the weapon/tool split and put hybrid into both tags so they could use modifier groups from either, and if we find a scenario where we want something specific for hybrid/axes it could be an option. We'll have to revisit how distinct weapons should handle modifiers at a later time, perhaps with more feedback.

Rev219:
Ok. Working on that now

Rev219:
Re-added the main hand tag and fixed the pull request issues, everything look good?

src/generated/resources/data/wotr/wotr/modifier/mining_attack_hybrid.json
ImplementsLegend, line 1: this still exists?

src/generated/resources/data/wotr/tags/item/socketable_main_hand_slot_weapon.json
```diff
@@ -0,0 +1,18 @@
...
+    "minecraft:diamond_sword",
+    "minecraft:netherite_sword",
+    "minecraft:bow",
+    "minecraft:crossbow",
+    "minecraft:wooden_axe",
+    "minecraft:stone_axe",
+    "minecraft:iron_axe",
+    "minecraft:golden_axe",
+    "minecraft:diamond_axe",
+    "minecraft:netherite_axe"
```
ImplementsLegend, line 16: We didn't ask you to completely remove the hybrid tag, only to not-give it any special modifiers but instead include it in the other tags (like how `#wotr:socketable` is made of other tags): 
```suggestion
    "#wotr:socketable_main_hand_slot_hybrid"
```
And where did the mace and trident go?

src/generated/resources/data/wotr/tags/item/socketable_main_hand_slot.json
ImplementsLegend, line 1: This should also be made of the other tags
```json
{
    "values": [ "#wotr:socketable_main_hand_slot_weapon", "#wotr:socketable_main_hand_slot_tool" ]
}
```

src/generated/resources/data/wotr/tags/item/socketable_main_hand_slot_weapon.json
Rev219: they weren't in there

src/generated/resources/data/wotr/tags/item/socketable_main_hand_slot.json
Rev219: image
so like this?

ImplementsLegend: Hybrid should already be part of both tools and weapons so there should be no need to add it to all mainhand. Other than that, yes.

PR #369, opened 2025-09-19
Anomalies

ImplementsLegend:
Anomaly block
====
Anomaly tasks
---
`data/wotr/wotr/anomaly_task`
Main task class for handling the logic and configuration. State class for tracking the task progress.
Tasks currently implemented:
- `NeedleTask`: right-click anomaly with specific item several times. (In previous PR required special "Needle" item.)
- `BundleTask`: right-click anomaly with bundle to scrifice it's content to the anomaly. (Wouldn't it be better to feed the items directly instead of from a bundle?)
- `BattleTask`: right-click anomaly to spawn mobs that need to be killed

Anomaly rewards
---
`data/wotr/wotr/anomaly_reward`
Reward may include:
- Effect to be given to player that completed the anomaly.
- Loot to be given to player that completed the anomaly.

Anomaly Structure Processor
====
Has weighted lists for task and reward selection.

Anomaly Generation
====
15% chance that anomaly will replace 3×3 POI


src/main/java/com/wanderersoftherift/wotr/block/blockentity/AnomalyBlockEntity.java
```diff
@@ -0,0 +1,230 @@
...
+            }
+        }
+    }
+
+    public void scheduledTick(ServerLevel serverLevel, BlockPos pos, BlockState state1) {
+        if (state == null) {
+            return;
+        }
+        var taskState = state.state();
+        if (taskState.isPresent() && taskState.get() instanceof BattleTaskState battleTask) {
```
immortius, line 75: Feels odd to have BattleTaskState behavior hardcoded here, maybe should have a tick method in taskState and delegate to that.

src/main/java/com/wanderersoftherift/wotr/block/blockentity/anomaly/AnomalyReward.java
```diff
@@ -0,0 +1,49 @@
...
+import net.minecraft.server.level.ServerLevel;
+import net.minecraft.util.context.ContextKeySet;
+import net.minecraft.world.effect.MobEffectInstance;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.level.storage.loot.LootParams;
+import net.minecraft.world.level.storage.loot.LootTable;
+
+import java.util.Optional;
+
+public record AnomalyReward(Optional<MobEffectInstance> effect/* potential alternative: it could cast an ability */,
```
immortius, line 19: Might be worth combining with the quest reward data type in the future.

src/main/java/com/wanderersoftherift/wotr/block/blockentity/AnomalyBlockEntity.java
```diff
@@ -0,0 +1,230 @@
...
+                    closeAndReward(player);
+                    return;
+                }
+            }
+        }
+        serverLevel.scheduleTick(pos, state1.getBlock(), 1);
+
+    }
+
+    public void battleMobDeath(LivingEntity entity) {
```
immortius, line 88: Likewise this BattleEvents could probably handle this logic.

immortius (approved):
Left a couple of comments, but given this is a PoC and been worked on for ages I think it is fine to add as is and improve in the future if this is a feature we want to carry forwards after testing.

PR #375, opened 2025-10-03
Enchantments modifiers

ImplementsLegend:
Ports enchantments to our modifier system.

There are two versions, so first we should decide which version to use:

Version A (this PR): Uses `GetEnchantmentLevelEvent` to check which modifiers are present on a tool. Advantage: it's really simple. Disadvantage: these modifiers will only work on equipment the enchantments are intended for (e.g. silk touch only works on tools, doesn't work on armor), which contradicts how other modifiers work (attack damage works whether it's on sword, shield or armor)

Version B (the other PR): enchantment modifiers register their enchants to player attachment. Logic using those enchantments is then modified to allow reading from this attachment. Advantage: enchantment modifiers are no longer limited to specific equipment. Disadvantage: this method requires individually adjusting all uses of enchantments - a task of quite unknown magnitude. Currently only ports loot tables. This required 3 mixins and even that might have missed some cases.


immortius:
Right. The choice is between a simpler implementation that has limitations in line with how enchantments work (silk touch has to be on the tool), or a much more involved implementation that removes this limitation, allowing silk touch on other equipment, or as an ability effect.

immortius (approved):

PR #381, opened 2025-10-07
Jyutta/itemobjective

jyutta:


src/main/java/com/wanderersoftherift/wotr/rift/objective/definition/CollectObjective.java
```diff
@@ -0,0 +1,91 @@
...
+
+    public static final MapCodec<CollectObjective> CODEC = RecordCodecBuilder.mapCodec(inst -> inst.group(
+            ExtraCodecs.POSITIVE_INT.fieldOf("min_quantity").forGetter(CollectObjective::getMinQuantity),
+            ExtraCodecs.POSITIVE_INT.fieldOf("max_quantity").forGetter(CollectObjective::getMaxQuantity),
+
+            BuiltInRegistries.ITEM.byNameCodec()
+                    .optionalFieldOf("item")
+                    .forGetter(o -> Optional.ofNullable(o.getItem())),
+
+            ResourceLocation.CODEC.optionalFieldOf("loot_table")
```
immortius, line 31: Can use `ResourceKey.codec(Registries.LOOT_TABLE)` here, and skip the need to convert the resource location.

src/main/java/com/wanderersoftherift/wotr/rift/objective/definition/CollectObjective.java
```diff
@@ -0,0 +1,91 @@
...
+import net.minecraft.world.level.ServerLevelAccessor;
+import net.minecraft.world.level.storage.loot.LootTable;
+
+import java.util.Optional;
+
+public class CollectObjective implements ObjectiveType {
+
+    public static final MapCodec<CollectObjective> CODEC = RecordCodecBuilder.mapCodec(inst -> inst.group(
+            ExtraCodecs.POSITIVE_INT.fieldOf("min_quantity").forGetter(CollectObjective::getMinQuantity),
+            ExtraCodecs.POSITIVE_INT.fieldOf("max_quantity").forGetter(CollectObjective::getMaxQuantity),
```
immortius, line 25: Might want to consider the possibility of the data author providing max_quantity < min_quantity. Although presumable should use the RiftParameter system that was just added.

src/main/java/com/wanderersoftherift/wotr/rift/objective/definition/CollectObjective.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public MapCodec<? extends ObjectiveType> getCodec() {
+        return CODEC;
+    }
+
+    @Override
+    public OngoingObjective generate(ServerLevelAccessor level, RiftConfig config) {
+        int target = level.getRandom().nextIntBetweenInclusive(minQuantity, maxQuantity);
```
immortius, line 64: It might be interesting to instead roll the loot generator X times, combine the result, and pick a random stackable item as the goal with the count equal to the result from all the rolls. This would build in accounting for rarity of different items, and it might be easier to balance around the expected number of chests to be opened.

src/main/java/com/wanderersoftherift/wotr/rift/objective/ongoing/CollectOngoingObjective.java
```diff
@@ -0,0 +1,243 @@
...
+        );
+    }
+
+    public boolean onInventoryCheck(Player player) {
+        if (isComplete()) {
+            return false;
+        }
+
+        int newAmount;
+        if (player instanceof ServerPlayer sp) {
```
immortius, line 97: Have you considered mutiplayer, with multiple players in the rift? Objectives are a shared state which will complicate things.

jyutta: Having trouble with the multiplayer part, can't even test if stuff works or not as when testing with two clients the other one crashes when the other enters a rift.

But other stuff should be done

immortius: Huh, interesting. I'll have to try and reproduce the crash.

Currently objectives are all shared. Not to say that cannot change in the future.

jyutta:
Not sure what to do with this, someone else should probably pick this up.

Issue #295, opened 2025-07-27
[Feature] New Fluids


WarrenGore:
Add new fluids, starting with a lavaux (fake lava, basically colored water). Expand to potentially have "harmful" and "harmless" fluids for themes to use in builds appropriately and keep with the theme/palette.

PR #402, opened 2025-10-14
Zleub/texture provider

Zleub:
Hello, this is a new provider for the mod, texture provider. The aim is to allow quick texture derivation as shown for processor blocks and water texture. 

This is a proposition to get #295 out of the way without tweaking anything from a FluidBlock.

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/init/client/ClientInitEvents.java
+++ b/src/main/java/com/wanderersoftherift/wotr/init/client/ClientInitEvents.java
@@ -105,7 +105,7 @@ public final class ClientInitEvents {
 
             @Override
             public ResourceLocation getOverlayTexture() {
-                return fluid.fluidOverlayId;
+                return null;
             }
 
             @Override


build.gradle
```diff
@@ -134,6 +134,9 @@ neoForge {
 // Include resources generated by data generators.
 sourceSets.main.resources { srcDir 'src/generated/resources' }
 
+processResources {
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
```
immortius: Are we expecting duplicates in the long term?

src/main/java/com/wanderersoftherift/wotr/datagen/textures/TintTransform.java
```diff
@@ -0,0 +1,28 @@
+package com.wanderersoftherift.wotr.datagen.textures;
+
+import net.minecraft.resources.ResourceLocation;
+
+import java.awt.Color;
+
+public class TintTransform extends TextureTransform {
+    Color tint;
```
immortius: Could be private.

src/main/java/com/wanderersoftherift/wotr/datagen/textures/TextureGeneratorCollector.java
```diff
@@ -0,0 +1,105 @@
...
+
+        ByteArrayOutputStream bytearrayoutputstream = new ByteArrayOutputStream();
+        HashingOutputStream hashingoutputstream = new HashingOutputStream(Hashing.sha1(), bytearrayoutputstream);
+
+        try {
+            Resource input = resourceManager.getResource(resourceLocation).orElseThrow();
+            ResourceMetadata resourceMeta = input.metadata();
+            AnimationMetadataSection textureMeta = resourceMeta.getSection(AnimationMetadataSection.TYPE).orElse(null);
+
+            InputStream r = input.open();
```
immortius: It would be good to ensure this is closed after use, either via a try-with-resources or an explicit close in a finally block.

immortius:
Oops, just realised this was still in draft, sorry >_<

src/main/java/com/wanderersoftherift/wotr/datagen/textures/TextureGenerator.java
ImplementsLegend: This class seems to have no purpose at all

build.gradle
Zleub: So, I got this added at the point where I had both textures, the original and the generated one. It seems that gradle copies over the generated ones when running the game and fail if no duplicate strategy is in place. There's only one discord message from mcjty that address this issue so I feel like we should keep it somehow even commented out 

Idk how you want to handle that, let me know

src/main/java/com/wanderersoftherift/wotr/datagen/textures/TextureTransform.java
```diff
@@ -0,0 +1,15 @@
+package com.wanderersoftherift.wotr.datagen.textures;
+
+import net.minecraft.resources.ResourceLocation;
+
+public abstract class TextureTransform {
+    public ResourceLocation sourcePath;
+    public ResourceLocation destinationPath;
+
+    TextureTransform(ResourceLocation sourcePath, ResourceLocation destinationPath) {
```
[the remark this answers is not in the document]
Zleub: Is there any reason to have those derived in another place than the `datagen\texture` folder ? If no then package private is fine no ?

WarrenGore:
@Zleub Is there anything more you were looking to do with this? I see it is still a draft.

immortius:
Could do with some language strings for the liquid buckets and blocks.

cptflavio:
There is a rendering issue for fluids on transparant block. 
image
(Middle is normal water)

I do think there is a use for both biome water color change (for waterlogging) and this custom fluid: to have several colored fluids at the same time. So both are nice!

PR #408, opened 2025-10-18
optimization: reduce number of particles created by anomalies

ImplementsLegend:
Anomaly particles were causing some frame-time stuttering. This reduces the number of particles based on distance from player.

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/block/blockentity/AnomalyBlockEntity.java
+++ b/src/main/java/com/wanderersoftherift/wotr/block/blockentity/AnomalyBlockEntity.java
@@ -60,19 +60,20 @@ public class AnomalyBlockEntity extends BlockEntity implements MobDeathNotifiabl
 
             var playerDistance = pos.distToCenterSqr(Minecraft.getInstance().player.position());
 
-            var count = PARTICLE_LODS.floorEntry(playerDistance).getValue();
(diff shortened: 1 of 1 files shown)


src/main/java/com/wanderersoftherift/wotr/block/blockentity/AnomalyBlockEntity.java
```diff
@@ -49,8 +58,14 @@ public void clientTick(ClientLevel clientLevel, BlockPos pos, BlockState state1)
...
-            var count = 6;
-            for (int i = 0; i < count; i++) {
+            var playerDistance = pos.distToCenterSqr(Minecraft.getInstance().player.position());
+
+            var count = PARTICLE_LODS.floorEntry(playerDistance).getValue();
+            if (count <= 0) {
+                return;
+            }
+            var delta = 6 / count;
+            for (int i = Math.floorMod(pos.hashCode(), delta); i < count; i += delta) {
```
immortius: I think this is double-applying the LOD reduction, by having both `i < count` and `i += delta`. For count = 2, delta would be 3, so you would only end up with 1 particles.

I assume the intent of delta was to spread the particles over the possible space, but count is used in the math below so I don't think it is necessary.

immortius (approved):

PR #417, opened 2025-10-25
Add custom tooltip rendering for Runegems

a1qs:
This PR adds a custom tooltip renderer for Runegems, displaying them neatly and adding pagination for all the possible socket types.
the modifier system has slightly changed since i last contributed but i think this should work properly for all the new stuff, at least i hope.

example:
image


src/main/java/com/wanderersoftherift/wotr/client/tooltip/RunegemTooltipRenderer.java
```diff
@@ -0,0 +1,353 @@
...
+            RunegemTooltipRenderer.maxIndex = -1;
+        }
+
+        @Override
+        public void onSlotClicked(@NotNull Slot slot, @NotNull ClickType clickType) {
+
+        }
+    }
+
+    public record RunegemComponent(ItemStack runegem, RunegemData data) implements TooltipComponent {
```
immortius: Looks like the ItemStack is only used to get the RunegemData so could probably remove it and just use the data.

src/main/java/com/wanderersoftherift/wotr/client/tooltip/RunegemTooltipRenderer.java
```diff
@@ -0,0 +1,353 @@
...
+
+    private static ResourceLocation getTierResourceLocation(RunegemTier tier) {
+        return WanderersOfTheRift.id("textures/tooltip/runegem/tier/" + tier.getName() + ".png");
+    }
+
+    private static ResourceLocation getShapeResourceLocation(RunegemShape shape) {
+        return WanderersOfTheRift.id("textures/tooltip/runegem/shape/text/" + shape.getName() + ".png");
+    }
+
+    public static class RunegemMouseActions implements ItemSlotMouseAction {
```
immortius: Perhaps pull this is to a top level class, as a singleton with currentIndex as a non-static member.

I could envision this being made generic rather than based on runegems specifically, probably with an interface that components implement to provide a page count. But something to revisit if that another use case for paginated tooltips arises.

src/main/java/com/wanderersoftherift/wotr/client/tooltip/RunegemTooltipRenderer.java
```diff
@@ -0,0 +1,353 @@
...
+import org.joml.Vector2i;
+
+import java.awt.Dimension;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+
+public class RunegemTooltipRenderer implements ClientTooltipComponent {
+    private static int currentIndex;
+    private static int maxIndex = -1;
```
immortius: Does this need to be static? Or possible could be derived as needed.

src/main/java/com/wanderersoftherift/wotr/client/tooltip/RunegemTooltipRenderer.java
```diff
@@ -0,0 +1,353 @@
...
+        for (TieredModifier tieredModifier : mods) {
+            Holder<Modifier> mod = tieredModifier.modifier();
+            int tier = tieredModifier.tier();
+
+            if (mod.getKey() == null) {
+                continue;
+            }
+
+            MutableComponent cmp = Component.literal("> ").withStyle(ChatFormatting.DARK_GRAY);
+            cmp.append(Component.literal("[T" + tier + "] ").withStyle(mod.value().getStyle()));
```
immortius: Potentially the `T` for tier should be translatable.

immortius (commented):
I'm noting there is also `socketable_main_hand_slot_weapon` (weapon) that is missing an tooltip info sprite currently.

src/main/java/com/wanderersoftherift/wotr/client/tooltip/RunegemTooltipRenderer.java
a1qs: yeah i was thinking of making it generic but couldnt really think of many cases where this would be needed, itd be easy enough to implement if we do find another usecase for it though.

src/main/java/com/wanderersoftherift/wotr/client/tooltip/RunegemTooltipRenderer.java
a1qs: missed that whoops

immortius (approved):

Issue #428, opened 2025-10-26
Quest configuration for hand-in location


immortius:
### Background

Currently once you take a quest, interacting with any quest giver will bring up the quest completion screen - so you can complete the quest at any quest giver. This both has the oddity of allowing you to hand in a quest at a different guild, and prevents you accepting multiple quests.

### Task

Add configuration for where a quest can be handed in. This could support options like:

* To the giver only
* To the giver's guild
* To a specific guild
* Entity Sub Predicate for fine grained control

Keeping in mind that (currently at least) the giver may be a block.

Attack8:
Test Comment :P

Attack8:
Another test

Attack8:
Test Comment

PR #439, opened 2025-11-01
(#299) Improve available quest handling to track per npc

### LINKED ISSUE #299: Link quests to guilds
Link quests to guilds so that guild quest givers know what quests to offer (and accept completion of). This should make use of tags so that it is easy to extend.

immortius:
The core of having guild specific quests was already done (quest giver npcs define a quest pool they pull from), however previously once the player had obtained a list of quests, that list was shown at all npcs, and the quest block. This change tracks separate lists of available quests per NPC. To do this it introduces the concept of an NpcIdentity, which identifies who an NPC. Both available quests and available trades are tracked against an NpcIdentity.

NpcIdentity is composed of three optional elements:
* A default entity type, which is used by a new command when creating the NPC.
* A guild the are a member of (currently unused)
* An NpcInteraction for when they are interacted with

Other changes:

* The QuestBlock has been replaced with an NpcBlock that can be tied to an NpcIdentity - a block for each NpcIdentity is automatically added to the creative menu.
* For backwards compatibility a default NpcIdentity exists (Bailey). This is just to bridge existing NPCs across (from the rift guild rooms).
* With the concept of NpcIdentity in place, quests have been extended to know which NpcIdentity they need to be handed in to - defaulting to the quest giver but with support for a different NPC if desired. The data for this supports either a single npc or a WeightedList of possible npcs.
* Existing quest menus and code has been adjusted to support having multiple quests, as you can now get one quest from each quest giver npc.
* To allow for quests to be handed in to non-quest giver npcs a QuestReceiverInteraction has been introduced that can wrap another NpcInteraction (so a merchant can both accept quests specifically for them, followed by opening the merchant menu). Quest Givers have this built in via inheritance.
* Added some additional UI types (PanelContainer, RowContainer, WidgetEntry) and support for spacing between elements in a ScrollContainerWidget. Increasingly getting an itch to do a UI framework cleanup, getting a better sense for things.

Refactors

* MobInteraction has be renamed to NpcInteraction. 
* The AvailableQuestsMenu now uses Neoforge's IMneuTypeExtension support for sending extra data instead of using a custom packet, simplifying it.
* Moved interaction types into subpackage of npc

Closes #299 and #428

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/commands/NPCCommands.java
+++ b/src/main/java/com/wanderersoftherift/wotr/commands/NPCCommands.java
@@ -58,7 +58,6 @@ public class NPCCommands extends BaseCommand {
 
     private int makeNpc(CommandContext<CommandSourceStack> ctx, Entity mob, Holder<NpcIdentity> npcIdentity) {
         mob.setData(WotrAttachments.NPC_IDENTITY, new NpcIdentity.Attachment(Optional.of(npcIdentity)));
-        mob.setData(WotrAttachments.NPC_INTERACT, npcIdentity.value().npcInteraction());
         mob.setCustomName(NpcIdentity.getDisplayName(npcIdentity));
         return Command.SINGLE_SUCCESS;
     }
--- a/src/main/java/com/wanderersoftherift/wotr/core/npc/NpcEvent.java
+++ b/src/main/java/com/wanderersoftherift/wotr/core/npc/NpcEvent.java
@@ -3,6 +3,7 @@ package com.wanderersoftherift.wotr.core.npc;
 import com.wanderersoftherift.wotr.gui.menu.ValidatingLevelAccess;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Holder;
+import net.minecraft.world.InteractionHand;
 import net.minecraft.world.InteractionResult;
 import net.minecraft.world.entity.Mob;
 import net.minecraft.world.entity.player.Player;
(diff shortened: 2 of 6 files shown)


src/main/java/com/wanderersoftherift/wotr/core/quest/QuestMenuHelper.java
ImplementsLegend: Maybe this should actually be `QuestMenuProvider` instead.

src/main/java/com/wanderersoftherift/wotr/serialization/GuavaCodecs.java
```diff
@@ -28,4 +33,24 @@ public static <K, V> Codec<ListMultimap<K, V>> unboundListMultimap(Codec<K> keyC
...
+    /**
+     * Codec for a guava ListMultimap, where the key is not string serializable.
+     * 
+     * @param keyCodec   Codec for serializing the key
+     * @param valueCodec Codec for serializing the value
+     * @return
+     * @param <K> The type of the key.
+     * @param <V> The type of the value.
+     */
+    public static <K, V> Codec<ListMultimap<K, V>> complexKeyListMultimap(Codec<K> keyCodec, Codec<V> valueCodec) {
```
ImplementsLegend: The `complexKey` part of the name is unclear without reading the javadoc.

immortius: `compoundKey` perhaps?

src/main/java/com/wanderersoftherift/wotr/commands/NPCCommands.java
```diff
@@ -38,69 +33,42 @@ public NPCCommands() {
...
         );
-    }
 
-    private int makeMerchant(CommandContext<CommandSourceStack> context, Entity mob, Holder<LootTable> trades) {
-        mob.setData(WotrAttachments.MOB_INTERACT, new MerchantInteract(trades.getKey()));
-        return Command.SINGLE_SUCCESS;
+        builder.then(
+                Commands.literal("create")
+                        .then(Commands.argument(NPC_ARG, ResourceArgument.resource(context, WotrRegistries.Keys.NPCS))
+                                .then(Commands.argument(LOCATION_ARG, BlockPosArgument.blockPos())
```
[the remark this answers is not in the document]
immortius: Sure, although I would note that the BlockPosArgument supports both tab completion to the block the player is looking at, and `~ ~ ~` to use the player's location.

src/main/resources/data/wotr/wotr/npc/cats_cradle_merchant.json
```diff
@@ -0,0 +1,11 @@
+{
+  "guild": "wotr:cats_cradle",
+  "interaction": {
+    "type": "wotr:quest_receiver",
```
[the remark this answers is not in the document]
immortius: I have added an OnInteract event and moved the quest receiver functionality into an event handler for that. It uses InteractionResult to allow for cancelling further interaction processing.

mal0ki:
Ah thank you! That makes more sense.

PR #447, opened 2025-11-03
443 runegem cleanup

WarrenGore:
Various runegem and modifier changes focused on realigning gem shape to rune modifier expectations, filling in for missing socketable type - gem shape combinations, allowing rune modifier group for all available runegem tiers, further fixing previously given feedback, and changing other data that did not look correct. This is not meant as a numeric balancing attempt.
 
- Move leech life to heart gem (gives mainhand slot a heart modifier)
- Adjust lowest roll so it can't be 0 for various modifiers (changed previously to .01, but some modifiers don't use that many decimal places)
- Adjust ability cooldown so it doesn't roll as increased cooldown (bug)
- Add triangle options for off hand and chestplate so every gear piece has at least one modifier for each gem shape (if my data is correct, does not help if you roll more sockets than modifiers for that shape)
- Move critical (pentagon) and thorns (circle) to other shapes so diamond is only for monster gems (give more options for other shapes as well)
- Added other pre-existing critical modifiers to the critical gem
- Set modifier groups across all runegem tiers (leech life, critical, mining, and thorns were only set up for raw)
- Add runegem for existing heal power modifiers (heart option available to tools that could make sense, and additional option for a few other categories)
- Moved mining speed from circle to pentagon so there would be something applicable to tools in that shape
- changed health hybrid to refer to health percent (probably more of a nitpick but I was already changing a bunch)
- Changed Projectile Count to be a flat 1 or 2 since partials wouldn't have an effect
- Changed Projectile Pierce to be a flat 1 or 2 as it was not high enough as it was to do anything. Trying this out though it may be too strong - maybe it should be a modifier only available at higher tiers and value of 1 but that can be balanced later

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/datagen/WotrRuneGemDataProvider.java
+++ b/src/main/java/com/wanderersoftherift/wotr/datagen/WotrRuneGemDataProvider.java
@@ -829,7 +829,6 @@ public class WotrRuneGemDataProvider {
                                         new TieredModifier(1, getModifier(lookup, "projectile_pierce")),
                                         new TieredModifier(1, getModifier(lookup, "projectile_count")),
                                         new TieredModifier(2, getModifier(lookup, "projectile_pierce")),
-                                        new TieredModifier(2, getModifier(lookup, "projectile_count")),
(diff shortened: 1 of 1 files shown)


immortius (approved):
One small thought.

PR #454, opened 2025-11-07
Updated bug report URL to the new project name.

ItsKiyoto:
The URL still had the old project name, so that was updated to Wanderers-Of-The-Rift.


immortius (approved):

immortius:
I see there was some chatter about changing this to get he url from the toml. Happy if that is either done here or spun out as a separate issue.

Issue #451, opened 2025-11-05
Gear loot table updates


mal0ki:
A few different ideas are being worked on for adjusting the gear and it's sockets in the loot pool (with tiers etc). 
Creating this issue, so I can attach my branch to it, but it also means others can try create their versions off this. 

I'll attach links to relevant conversations on discord too. 

@andydirk 's take for datapack https://discord.com/channels/1374772629298483202/1408960619960602624/1416907917516345467
my take in ideasubmission: https://discord.com/channels/1374772629298483202/1431044532400553984

If anyone starts another conversation for implementation, let me know and I'll update this here. 

Attach branches or PRs to this issue.

PR #465, opened 2025-11-12
451 Sockets in gear loot table rift tiers

mal0ki:
This PR is for if people like the changes provided in the datapack (see discord), so it's all here and ready to merge if it get accepted as a change in the loot table by the community. 

For #451


mal0ki:
Thank you. It seems it was all off by one tab because of some change that I don't believe i made. At least it's fixed now.

PR #476, opened 2025-11-14
475 fix: remove villager from Tavern, correct room specific processor usage.

WarrenGore:
Remove villager from Tavern
Correct village room 3 specific processor usage


mal0ki:
iirc the issue is in all the versions of village room by rachael, so 1-3 (don't recall if there's more), unless the other two have been fixed already.

mal0ki:
Okay, maybe that's what rachael referred to the other day too. 👍

PR #488, opened 2025-11-17
(#120) wotr loot commands

### LINKED ISSUE #120: [Commands] Mod Chest Creation
I would like to investigate an option that could create a rift chest outside the rift with a specified rift tier for testing loot tables. 
it could simply spawn at a block location or use the player's target.

It would make life much easier if you could simply spawn in a few chests and simply open them to test tweaks to loot tables. 

A nice to have would be to also be able to specify another param that would let you spawn them in a 3x3, 4x4, 5x5 etc (with rift tier of course)

immortius:
Closes #120 

This PR adds two commands:

`wotr loot createChest <location> <loot_table> [<rift_tier>]`
`wotr loot fillAreaWithChests <from> <to> <loot_table> [<rift_tier>]`

These commands place Rift Chests at the specific location(s) and generate their content from the provided loot table. The provided rift tier is used as a parameter for this generation. The commands can be rerun to effectively regenerate the chest content.

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/commands/LootCommands.java
+++ b/src/main/java/com/wanderersoftherift/wotr/commands/LootCommands.java
@@ -76,12 +76,10 @@ public class LootCommands extends BaseCommand {
             Holder<LootTable> lootTable,
             int riftTier) {
         BoundingBox area = BoundingBox.fromCorners(from, to);
-        int result = 0;
-        for (BlockPos pos : BlockPos.betweenClosed(area.minX(), area.minY(), area.minZ(), area.maxX(), area.maxY(),
-                area.maxZ())) {
-            result += createChest(ctx, pos, lootTable, riftTier);
-        }
-        return result;
+        return BlockPos
+                .betweenClosedStream(area.minX(), area.minY(), area.minZ(), area.maxX(), area.maxY(), area.maxZ())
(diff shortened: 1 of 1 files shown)


src/main/java/com/wanderersoftherift/wotr/commands/LootCommands.java
```diff
@@ -0,0 +1,109 @@
...
+    private int createChests(
+            CommandContext<CommandSourceStack> ctx,
+            BlockPos from,
+            BlockPos to,
+            Holder<LootTable> lootTable,
+            int riftTier) {
+        BoundingBox area = BoundingBox.fromCorners(from, to);
+        int result = 0;
+        for (BlockPos pos : BlockPos.betweenClosed(area.minX(), area.minY(), area.minZ(), area.maxX(), area.maxY(),
+                area.maxZ())) {
```
ImplementsLegend:
```suggestion
        int result = 0;
        for (BlockPos pos : BlockPos.betweenClosed(from, to)) {
```

Maybe you could even use stream
```java
return BlockPos.betweenClosedStream(from, to)
        .mapToInt(pos->createChest(ctx, pos, lootTable, riftTier))
        .sum();
```

ImplementsLegend (approved):

PR #490, opened 2025-11-20
(#424) Reset available quests command

### LINKED ISSUE #424: [Command] Add Available Quest List Reset Command
Quests currently only reset with an empty list or on rift exit. Add a new command that would generate a new list of available quests.

Something like /wotr quest resetList <target player>

immortius:
Additionally added a command to reset the quest log, and reworked the quest command structure. The possible commands are:

`wotr quest log list [<target player>]`
`wotr quest log reset [<target player>]`
`wotr quest available reset [<target player>]`

Closes #424


Attack8 (changes requested):
`quest active list` command, as discussed in vc on 2025.11.29

Attack8 (approved):

PR #495, opened 2025-11-21
Add: Discount Hookshot

cptflavio:
Incomplete hookshot. Should be projectile, and preferably a tether between source and target. When available, stun instead of slow.


cptflavio:
once polished, it should include some upgrades too

Issue #486, opened 2025-11-17
Add Essence As Items


WarrenGore:
Add essence items that can be found within rifts.

- Add the essences currently in use
  - They will need dev art
- Add essence items to loot tables
  - To start, just to wooden.json is fine, can adjust per theme later when we work those changes out
- Add essence value to the essences
  - For now, 10 essence of its respective type

Essence types in use today
- Earth
- Water
- Plant
- Death
- Light
- Nether
- Life
- Honey
- Mushroom
- Fabric
- Dark

Essence types yet to be used (don't need to be added at this time but feel free to if you are up to it)
- Fire
- Air
- Energy
- Animal
- Crystal
- Metal
- Food
- Slime
- Mind
- Mecha
- End
- Flow
- Form
- Order
- Chaos

PR #507, opened 2025-11-29
FEAT: #486 Add Essence As Items

family2015:
- Placeholder textures provided by @Danymaddox


src/main/java/com/wanderersoftherift/wotr/init/WotrItems.java
```diff
@@ -107,6 +109,65 @@ public class WotrItems {
...
+            EssenceType.FLOW.getName().toLowerCase() + "_essence");
+    public static final DeferredItem<Item> FORM_ESSENCE = registerEssence(
+            EssenceType.FORM.getName().toLowerCase() + "_essence");
+    public static final DeferredItem<Item> ORDER_ESSENCE = registerEssence(
+            EssenceType.ORDER.getName().toLowerCase() + "_essence");
+    public static final DeferredItem<Item> CHAOS_ESSENCE = registerEssence(
+            EssenceType.CHAOS.getName().toLowerCase() + "_essence");
+
+    public static final DeferredItem<Item> registerEssence(String essenceName) {
+        return ITEMS.register(essenceName, registryName -> new EssenceItem(
```
Attack8:
```suggestion
        return ITEMS.register(essenceName, registryName -> new Item(
```
Because the `EssenceItem` class is not needed, this can be changed to just be `Item`

src/main/java/com/wanderersoftherift/wotr/item/essence/EssenceItem.java
Attack8: This class can be removed as stacks to 64 is the default

src/main/java/com/wanderersoftherift/wotr/datagen/WotrChestLootTableProvider.java
```diff
@@ -70,10 +72,6 @@ public void generate(BiConsumer<ResourceKey<LootTable>, LootTable.Builder> consu
                                         .when(riftTier(2, 5))
                                         .setWeight(20)
                                         .apply(SetItemCountFunction.setCount(UniformGenerator.between(2.0F, 4.0F))))
-                                .add(LootItem.lootTableItem(WotrItems.SKILL_THREAD)
-                                        .when(riftTier().min(5))
-                                        .setWeight(20)
-                                        .apply(SetItemCountFunction.setCount(UniformGenerator.between(3.0F, 6.0F))))
```
Attack8: This should not have been removed as far as I know, please revert this change!

Attack8 (changes requested):

src/main/java/com/wanderersoftherift/wotr/item/essence/EssenceItem.java
```diff
@@ -0,0 +1,9 @@
+package com.wanderersoftherift.wotr.item.essence;
+
+import net.minecraft.world.item.Item;
+
+public class EssenceItem extends Item {
```
immortius: Could just use `Item` and not have a subclass.

src/main/java/com/wanderersoftherift/wotr/init/WotrItems.java
```diff
@@ -107,6 +109,65 @@ public class WotrItems {
     public static final DeferredItem<Item> COLOR_HELMET = ITEMS.register("color_helmet",
             () -> new WotrArmor(EquipmentSlot.HEAD, "color_helmet", 165));
 
+    // Essence
+    public static final DeferredItem<Item> EARTH_ESSENCE = registerEssence(
+            EssenceType.EARTH.getName().toLowerCase() + "_essence");
```
immortius: Please use `toLowerCase(Locale.ROOT)` - otherwise will break for people running with some languages (example: capital `I` lowercases to `ı` in Turkish). Could move the EssenceType -> id logic into the registerEssence method and just pass the enum.

src/main/java/com/wanderersoftherift/wotr/item/essence/EssenceType.java
```diff
@@ -0,0 +1,57 @@
+package com.wanderersoftherift.wotr.item.essence;
+
+import java.util.Arrays;
+
+public enum EssenceType {
```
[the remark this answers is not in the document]
mal0ki: I do think being able to datapack it could be good for the long-term. Especially if we're looking at testing different essence "layouts" etc long term.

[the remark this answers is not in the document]
Danymaddox: Im a bit confused: you mean create new essence types or change/add essences from/to items?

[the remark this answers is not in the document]
mal0ki: Good quesstion @Danymaddox ( I don't know what I meant, nothing specific, I just lvoe them datapacks right now, but I was thinking of all the different essence related suggestions there's been, there's been quite a few conversations).

src/main/java/com/wanderersoftherift/wotr/datagen/WotrChestLootTableProvider.java
mal0ki: Increase the weight of the rift/essences on line 58. 
5 is way too low for them to work in the loot pool right now. 
I think having it on par with the skill thread etc is fine.

src/main/java/com/wanderersoftherift/wotr/item/essence/EssenceType.java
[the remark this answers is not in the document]
immortius: The currency bag, for example, is a single item but has a data component that determines the type of currency it contains. And it uses the Emblem renderer we added to render the currency icon on top of a money bag item model. Abilities are the same - one base item which is configured by data component and renders the ability icon on the holder model. This means the currencies and abilities can be added via data without needing a new item for each one. So for essences would be having a single item and a data type for defining the possible types of essence.

But this is a significant piece of work, so can also add a new issue for it and leave it for a future PR, perhaps after people have experienced essence as an item and we're moving forward with it. Also the Emblem renderer might not be what we want in this case.

Attack8 (approved):

PR #538, opened 2025-12-19
#457 Inconsistent selection of RNG algorithm

### LINKED ISSUE #457: Inconsistent selection of RNG algorithm
WotR at a number of points creates random number generators from available Java RNG algorithms. This is done by producing a list of available algorithms, and then selecting from them by index - generally 0, but in one case 6. This provides no guarantee of which algorithm is being selected across java distributions as the available algorithms and order may be different - and in the case there are less than 7 algorithms available WotR will crash during rift generation.

I would suggest that in the cases using 0, the default algorithm should be used (`RandomGeneratorFactory.getDefault()`). For the case using 6, should attempt to use the specific algorithm intended and then fall back on the default if not available (or perhaps have a list of fallbacks ending on the default).

ItsKiyoto:
Closes #457

Implemented a Enum system to replace the index method of retrieving RNG algorithms, so previous calls for algorithm at index 0 now use the default algorithm and the instance where a specific algorithm was used (at index 6) is using the same algorithm but just referenced by name. 

The current implementation only allows for the fallback to be the default algorithm but that can be implemented in cases where it may be more necessary.

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/util/RandomFactoryType.java
+++ b/src/main/java/com/wanderersoftherift/wotr/util/RandomFactoryType.java
@@ -14,7 +14,6 @@ public enum RandomFactoryType {
         @Override
         RandomGeneratorFactory get() {
             return RandomGeneratorFactory.all()
-                    .filter((it) -> !it.isDeprecated())
                     .filter(f -> "Xoshiro256PlusPlus".equals(f.name()))
                     .findFirst()
                     .orElse(RandomGeneratorFactory.getDefault());


src/main/java/com/wanderersoftherift/wotr/util/RandomSourceFromJavaRandom.java
```diff
@@ -61,8 +57,8 @@ public void parityConfigString(StringBuilder stringBuilder) {
         };
     }
 
-    public static RandomGeneratorFactory get(int i) {
-        return GENERATORS.get(i);
+    public static RandomGeneratorFactory get(RandomFactoryType factoryType) {
+        return factoryType.get();
```
immortius: We do lose the caching of factory types with this approach... perhaps there could be an EnumMap populated in a static block. Though I don't know if it matters too much - if we only occasionally create new generators factories or they aren't expensive to create then it doesn't.

PR #542, opened 2025-12-28
Experimental dungeon layout for rifts

immortius:
### Overview

Introduces an experimental dungeon layout for rifts. Places rooms as branches, with major rooms separated by connecting rooms. By default there is a major room every 3 rooms along the branches, and at a major room there is a 50% chance of a second branch.

### Usage

To test, modify a key with the command: `/wotr riftKey generator setPreset wotr:dungeon`. The size of the rift is currently determine by tier, with no limit although 16 seems to be plenty. At some size performance will tank/memory will be exhausted, but this is true for large finite rift layouts too.

The generation can be further finetuned with:
* `/wotr riftKey parameter wotr:dungeon_rift/branch_rate <value>` - how many branches from each main room, with fractional parts being the chance for an additional branch. Defaults to `1.5`
* `/wotr riftKey parameter wotr:dungeon_rift/main_room_interval <value>` how frequently main rooms (3x_x3) appear. Main rooms appear immediately after the portal room, and then every nth room as determined by this value. Defaults to `3`.
* `/wotr riftKey parameter wotr:dungeon_rift/max_depth <value>` how many rooms out from the portal room the branches will generate. Defaults to the tier of the rift.

### Future Work

There is undoubtedly a lot that can be done to expand or improve this approach. For example, adding terminator rooms, improving the room pools, better randomisation of branching, optimisation of generation. The logic behind selecting an placing a room is fairly basic - more could be done here though it will add complexity and cost. Gameplay-wise having objectives inserting only into main rooms may be interesting. Having exit portals or teleporters between terminating rooms may also help prevent backtracking if that is an issue too.

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/core/rift/parameter/definitions/ConstantRiftParameter.java
+++ b/src/main/java/com/wanderersoftherift/wotr/core/rift/parameter/definitions/ConstantRiftParameter.java
@@ -6,7 +6,7 @@ import net.minecraft.util.RandomSource;
 
 import java.util.function.Function;
 
-public record ConstantRiftParameter(double value) implements RiftParameter {
+public record ConstantRiftParameter(double value) implements RiftParameterDefinition {
     public static final Codec<ConstantRiftParameter> CODEC = Codec.DOUBLE.xmap(ConstantRiftParameter::new,
             ConstantRiftParameter::value);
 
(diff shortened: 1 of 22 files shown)


ImplementsLegend:
There's no vertical limit to the placement of the rooms so rooms can break through bedrock.
image
In other layouts this is prevented using `RiftShape`, maybe consider adding one.

There is also no simple way of explicitly adding specific rooms which could normally just be their own `LayoutLayer`.

Is there a reason why this can't be a layout layer instead? That would solve both problems.

I guess there might be some benefit of storing the rooms in a hashmap (to conserve memory) instead of a larga array as the other layouts do so maybe this should be split into `SparseLayeredLayout` and `BranchingLayoutLayer`.

PR #548, opened 2026-01-28
Feat: Overworld Guilds (+ability update)

cptflavio:
Spawns guilds in the overworld, just like small villages. Inside those guilds are merchants and quest givers. 
This is a placeholder for hubworld, not final. 

Also updated breath abilities to actually be in front of player, not around.
And a few new abilities: testing step-up and scale attributes.


mal0ki:
Can you try to not mix and match different thigns that are entirely unrelated in the same PR? (I know it's not a big deal here, but just make a habit of it to keep it seperate)

PR #551, opened 2026-01-30
Anomaly themes

### LINKED ISSUE #382: Anomoly Mobs Spawn in invalid locations
Anomoly Mobs can spawn with their heads in blocks (causing them to suffocate) or get their AI stuck on the other blocks in the POIs. They should search for a nearby safe location to spawn in.

### LINKED ISSUE #528: Anomalies Set By "Origin" Theme
Anomalies have been described as being connections to other rifts. To align with this, allow anomalies to pull from a random theme, and once chosen for that anomaly pull that theme's:
- palette of blocks for POI processing
- loot table for rewards
  - we don't have themed loot tables created yet, but allow either a loot table to be defined per theme or read the loot table from the poi_theme_<themeName> file under chest_types (keep https://discord.com/channels/1374772629298483202/1439338634397483198 in mind in case chest types are added before this)
- if anomaly spawns mobs, mobs should match the anomaly theme

It is okay if the random theme happens to be the current rift's theme, does not need to be exclusive.

ImplementsLegend:
Adds "origin" themes to anomalies (closes #528)
- each theme has different palette
- each theme requires different essence types for essence task
- each theme has different mobs for battle task

Refactored spawn functions
- moved to separate package
- added more spawn functions, including new spawn function to search for spawn location (closes #382)
- added spawn function datapack registry

Future work
- The way anomaly themes are currently applied is suboptimal since it creates many duplicate entries in template pools. This could be addressed by refactoring `JigsawListProcessor` in a way that allows it to make more modifications to the substructures being generated like changing theme.
- some themes have special dependencies - what is the best way to enable those?


ImplementsLegend:
I made that change since I think it's more useful to make adding new themes easier. Probably ok to merge now.

PR #555, opened 2026-01-31
Changed anomalies to always give a reward

family2015:
This PR aims to do the next things:
- Change the default rewards to have a loot table and an effect.
- Add new loot tables for the different rift themes.

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/datagen/DataGenerators.java
+++ b/src/main/java/com/wanderersoftherift/wotr/datagen/DataGenerators.java
@@ -69,7 +69,7 @@ public class DataGenerators {
                 new LootTableProvider.SubProviderEntry(WotrBlockLootTableProvider::new, LootContextParamSets.BLOCK),
                 new LootTableProvider.SubProviderEntry(WotrChestLootTableProvider::new, LootContextParamSets.CHEST),
                 new LootTableProvider.SubProviderEntry(WotrAnomalyRewardLootTableProvider::new,
-                        LootContextParamSets.COMMAND),
+                        LootContextParamSets.EMPTY),
                 new LootTableProvider.SubProviderEntry(WotrRiftObjectiveLootTableProvider::new,
                         LootContextParamSets.EMPTY),
                 new LootTableProvider.SubProviderEntry(WotrLootBoxLootTableProvider::new, LootContextParamSets.EMPTY)),


src/main/java/com/wanderersoftherift/wotr/block/blockentity/anomaly/AnomalyReward.java
```diff
@@ -14,16 +14,12 @@
 import net.minecraft.world.level.storage.loot.LootParams;
 import net.minecraft.world.level.storage.loot.LootTable;
 
-import java.util.Optional;
-
-public record AnomalyReward(Optional<MobEffectInstance> effect/* potential alternative: it could cast an ability */,
-        Optional<ResourceKey<LootTable>> lootKey) {
+public record AnomalyReward(MobEffectInstance effect/* potential alternative: it could cast an ability */,
```
immortius: I don't think these code changes are necessary - and it would be good to keep the two possible reward types as optional is it isn't necessary for an Anomaly to reward both. Changing the configuration to have a guaranteed reward is probably enough (although I would note that the loot table is still only a chance of giving an item).

immortius (changes requested):
The anomaly reward data fix is good, although there's still only a chance of getting an item. I was wondering if there is a reason that strength has been replaced with absorption?

family2015:
> @family2015 With #551 merged is there anything more you need to move this forward?

Didn't notice it was merged, I will start to work on it

family2015:
Right now my game crashes when I try to create or join a world, I not sure on what is wrong
[latest.log](https://github.com/user-attachments/files/25400640/latest.log)

src/main/java/com/wanderersoftherift/wotr/datagen/DataGenerators.java
```diff
@@ -68,6 +68,8 @@ public static void gatherData(GatherDataEvent.Client event) {
         event.createProvider((output, lookupProvider) -> new LootTableProvider(output, Set.of(), List.of(
                 new LootTableProvider.SubProviderEntry(WotrBlockLootTableProvider::new, LootContextParamSets.BLOCK),
                 new LootTableProvider.SubProviderEntry(WotrChestLootTableProvider::new, LootContextParamSets.CHEST),
+                new LootTableProvider.SubProviderEntry(WotrAnomalyRewardLootTableProvider::new,
+                        LootContextParamSets.COMMAND),
```
immortius: I would probably use `LootContextParamSets.EMPTY` to match what AnomalyReward is configuring the loot context with, unless there is some other intent here. Maybe we should make our own LootContextParamSet specifically for Anomalies.

PR #561, opened 2026-02-03
Dev Environment Cleanup

Attack8:
Update mods in dev env to reflect the packs being distributed and add motr as a dependency


immortius (approved):
This looks good to me.

PR #562, opened 2026-02-03
Feat: Cuboid Targeting Area

### LINKED ISSUE #503: [Abilities] Area target Cuboid
Rather than a cube, it would be nice to have a cuboid, which accepts X Y and Z lenghts. Useful for something like a wall spell.

cptflavio:
Adds a cuboid shape to area targeting.
Adds UseOriginDirection boolean to keep the direction of source prior to Area Targeting, such that walls can be created depending on facing of player, rather than the direction of first target hit.

Closes #503


cptflavio:
> There may be an argument for dropping cube shape and just having a different codec for cube and cuboid.
> 
> Otherwise looks good, make sure to update the ability doco in wotr-docs too.

I think its good to keep a range for simplicity, and cube/cuboid makes the difference clear, but I do see a use in fusing them. A bit on the fence, dont mind changing later.

Ill update the docs now!

PR #563, opened 2026-02-03
Fix: NoMoveRequirement

### LINKED ISSUE #506: [Abilities] Bug: No_Move broken
`NoMoveRequirement.java` currently does not work: X and Y always output 0.0. 
Delta probably doesnt get the correct values. There is a Y value, but that always prints -0.078, even when stationary.
Values gained from adding a print(delta) in the java file.

The JSON I tested with:
```json
{
  "type": "wotr:persistent",
  "activation_effects": [
    {
      "type": "wotr:attach",
      "id": "wotr:sneaky_sneak",
      "trigger": {
        "frequency": 1
      },
      "effects": [
        {
          "type": "wotr:status",
          "status_effect": {
            "id": "minecraft:invisibility",
            "amplifier": 10,
            "duration": 20
          }
        }
        ]
    }
  ],
  "on_deactivation_effects": [
    {
      "type": "wotr:detach_own",
      "id": "wotr:sneaky_sneak"
    }
  ],
  "ongoing_requirements": [
    {
      "type": "wotr:no_move",
      "max_speed": 0.0001
    }
  ],
  "requirements": [
    {
      "type": "wotr:cooldown",
      "ticks": 10
    },
    {
      "type": "wotr:ability_resource",
      "resource_type": "wotr:mana",
      "amount": 1
    }
  ],
  "icon": "minecraft:textures/mob_effect/invisibility.png"
}
```

cptflavio:
Fixes No Move.
Adds an optional include_y, to add y-velocity to allowed speed.
Slight reading change on CuboidShape, remove leftover from testing. 

Closes #506

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/abilities/requirement/NoMoveRequirement.java
+++ b/src/main/java/com/wanderersoftherift/wotr/abilities/requirement/NoMoveRequirement.java
@@ -30,6 +30,6 @@ public record NoMoveRequirement(float maxSpeed, boolean includeY) implements Abi
         Vec3 delta = new Vec3(
                 context.caster().xo - context.caster().xOld, dY, context.caster().zo - context.caster().zOld
         );
-        return delta.length() < this.maxSpeed;
+        return delta.lengthSqr() < (this.maxSpeed * this.maxSpeed);
     }
 }


cptflavio:
Previously used .getDeltaMovement, which does not work server side. getX() will cause problems with lag.

src/main/java/com/wanderersoftherift/wotr/abilities/requirement/NoMoveRequirement.java
```diff
@@ -19,11 +20,16 @@ public MapCodec<? extends AbilityRequirement> getCodec() {
...
     public boolean check(AbilityContext context) {
-        Vec3 delta = context.caster().getDeltaMovement();
-        if (context.caster().onGround()) {
-            delta = delta.multiply(1, 0, 1);
+        if (context.level().isClientSide()) {
+            return true;
         }
 
-        return delta.lengthSqr() < maxSpeed * maxSpeed;
+        double dY;
```
immortius: This could be combined with the next line - I assume it was separate because at one stage a full if/else was being used?

PR #567, opened 2026-02-05
Attackable Blocks

### LINKED ISSUE #335: Destroyable Rift Chests
Add support for destroying rift chests with straight damage (e.g. weapon attacks, abilities) rather than needing to open them or harvest them with a tool.

ImplementsLegend:
Some blocks can now be attacked with abilities, including weapon abilities. Those blocks are: rift chest, decorated pot, bamboo, leaves, cobweb, "bushes" and some types of anomalies (new anomaly task). Chests and anomalies have HP, the rest breaks in one hit.

Future work:
- Chests could have dynamic HP based on content and drop part of it's content every hit.
- The anomaly task could check the type of damage.

closes #335

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/init/WotrTags.java
+++ b/src/main/java/com/wanderersoftherift/wotr/init/WotrTags.java
@@ -16,6 +16,7 @@ public class WotrTags {
 
     public static class Blocks {
         public static final TagKey<Block> BANNED_IN_RIFT = createTag("banned_in_rift");
+        public static final TagKey<Block> ATTACKABLE = createTag("attackable");
 
         private static TagKey<Block> createTag(String name) {
             return BlockTags.create(WanderersOfTheRift.id(name));
--- a/src/main/java/com/wanderersoftherift/wotr/mixin/MixinBlock.java
+++ b/src/main/java/com/wanderersoftherift/wotr/mixin/MixinBlock.java
@@ -1,23 +1,20 @@
 package com.wanderersoftherift.wotr.mixin;
 
 import com.wanderersoftherift.wotr.block.AttackableBlock;
+import com.wanderersoftherift.wotr.init.WotrTags;
 import net.minecraft.core.BlockPos;
 import net.minecraft.world.damagesource.DamageSource;
 import net.minecraft.world.level.BlockGetter;
 import net.minecraft.world.level.LevelAccessor;
-import net.minecraft.world.level.block.BambooStalkBlock;
-import net.minecraft.world.level.block.BushBlock;
-import net.minecraft.world.level.block.DecoratedPotBlock;
-import net.minecraft.world.level.block.LeavesBlock;
-import net.minecraft.world.level.block.WebBlock;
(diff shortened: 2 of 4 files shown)


src/main/java/com/wanderersoftherift/wotr/abilities/effects/predicate/TargetBlockPredicate.java
```diff
@@ -66,7 +67,14 @@ public boolean matches(BlockPos target, HitResult source, AbilityContext context
 
     enum Trivial implements TargetBlockPredicate, StringRepresentable {
         ALL("all", true),
-        NONE("none", false);
+        NONE("none", false),
+        ATTACKABLE("attackable", true) {
+            @Override
+            public boolean matches(BlockPos target, HitResult source, AbilityContext context) {
```
immortius: Also could remove the boolean value from the constructor and have matches abstract and set per entry. Happy either way.

src/main/java/com/wanderersoftherift/wotr/block/AttackableBlock.java
```diff
@@ -0,0 +1,42 @@
...
+                return super.getBlockShape(blockState, level, pos);
+            }
+        };
+    }
+
+    static boolean isAttackableStatic(BlockState state, BlockGetter level, BlockPos position) {
+        if (state.getBlock() instanceof AttackableBlock ab && ab.isAttackable(state, level, position)) {
+            return true;
+        }
+        if (level.getBlockEntity(position) instanceof AttackableBlock ab && ab.isAttackable(state, level, position)) {
```
immortius: There might be value in only checking for blocks being AttackableBlock, and then having the block types delegate to their entity when it is appropriate. Would save lookups for block entities where blocks cannot have one.

src/main/java/com/wanderersoftherift/wotr/block/blockentity/AnomalyBlockEntity.java
```diff
@@ -226,6 +232,26 @@ public void setPanorama(ResourceLocation resourceLocation) {
         this.panorama = resourceLocation;
     }
 
+    @Override
+    public boolean isAttackable(BlockState state, BlockGetter level, BlockPos position) {
+        var anomalyState = this.getAnomalyState();
+        return anomalyState != null && anomalyState.state.isPresent()
+                && anomalyState.task.value() instanceof SliceAnomalyTask;
```
immortius: Potentially AnomalyTask should have methods for isAttackable/attack that are delegated to, rather than checking the type. But admittedly hard to imagine another task type that cares about attacks.

As a side thought for future reflection, I wonder if it would be simpler just to have different blocks for different anomaly types, built on a common base. But we have another issue in that space.

src/main/java/com/wanderersoftherift/wotr/block/AttackableBlock.java
ImplementsLegend: I'll do that.

PR #569, opened 2026-02-06
Fix: Mana bar render & regen bugs

### LINKED ISSUE #509: Mana Bar Weirdness
The mana bar can sometimes render progress below the bar frame (Fig. 1) or mana can continue to be used below what is shown on the bar (Fig. 2). Additionally, when equipping gear with + max mana, the mana doesn't get regenerated until you use an ability, then it regenerates all the way. 

Image

Fig. 1

https://github.com/user-attachments/assets/9390e1ac-e064-4aa5-aa69-62ca225704f7
Fig. 2

cptflavio:
- Mana bar renders correctly now when mana is not multiple of 10. Used to dip below the bar frame. 
 - Mana regenerates after increasing max mana. This didnt happen when you had full mana, and increased the max. 

Closes #509

### DIFF
--- a/src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilityResourceData.java
+++ b/src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilityResourceData.java
@@ -147,7 +147,7 @@ public class AbilityResourceData {
 
         for (var entry : amounts.entrySet()) {
             if (entry.getKey().value().maximum().equals(attribute)) {
-                entry.getValue().onMaxCharged();
+                entry.getValue().onMaxChanged();
             }
         }
     }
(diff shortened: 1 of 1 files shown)


cptflavio:
The multiple resource render is not a perfect solution, and should be re-done. 
It used to render additional resources towards the left of the first, causing them to render off-screen, and outside the bounding box generated for them. This flips the direction of placement depending on position of screen. 

A better solution would be to remove the anchoring change within the bounding box, and make the bounding box more tight around the resource bars (currently there is some room on the right side). The resource bars can always start placement from top-left of their combined bounding box, as this increases when adding more resources.

src/main/java/com/wanderersoftherift/wotr/abilities/attachment/AbilityResourceData.java
```diff
@@ -186,6 +192,17 @@ void setHasNonZeroDelta(String eventKey, boolean newValue) {
             event.updateTriggers(TriggerTracker.forEntity((Entity) holder));
         }
 
+        void onMaxCharged() {
```
immortius: Should this be called `onMaxChanged` :) ?

immortius: I don't see any code calling this when the value is full - this is a new method only called when the attribute is changed, right?

PR #573, opened 2026-02-08
Quest Update

cptflavio:
- gives more coin rewards, to be used for merchants
 - Accepts RiftSkeleton towards kill_skeleton quest
 - Kill objective only accepts enemy mobs (used to count mobs like chickens too)


src/main/resources/data/wotr/wotr/objective/kill.json
```diff
@@ -3,6 +3,7 @@
   "goals": [
     {
       "type": "wotr:kill_mob",
+      "mob": "#wotr:rift_mobs",
```
immortius: Might want a mob_label as well, although I guess it is tricky since some of the mobs can be found outside the rift.

PR #577, opened 2026-02-10
Structure update

cptflavio:
- Adds new & updated structures, nbt and pool
 - Removes some broken files
 - Removes "podzol" error (finally)
 - Added rotation check before apply, so empty entity lists dont crash


cptflavio:
To check, perhaps run a few rifts or something, and check `TripleMirror.java`
Lots of changes are due to some re-structuring of documentation, which moved the order around.

PR #583, opened 2026-02-19
Instant loot framework

### LINKED ISSUE #581: Add Instant Loot
Add instant use loot
- Loot is consumed upon pickup - it never enters the inventory
  - Examples: Hearts for instant healing, mana vial for instant mana refill, food for instant stamina/hunger bar refill
  - May work similarly to vanilla experience orbs
- Has an area of effect only impacting players and player-aligned entities
- Can only be consumed by a player
  - mobs cannot trigger their effects

Based on loot category/source outline, should come from completed spawners.

ImplementsLegend:
Adds instant loot Data Component.

Any item with this data component will be consumed instantly when:
- picked up
- clicked in an inventory

Healing potions have been changed to an instant loot.

Closes #581 

Future work: 
Apply in AOE. (or cast an ability when instant loot is collected)
Add some new items with custom models.

Could be used for an alternate Item Objective.


immortius (approved):
Looks good. Might want to add InstantLoot::tryConsume to some of the methods in ItemStackHandlerUtil which are used when players close a temporary inventory screen (like the reward screen) so they don't end up with InstantLoot items in their inventory from that mechanism.

